PR #8, opened 2021-08-02
include aqt compilation functionality in cirq-superstaq

singular-value:


cirq_superstaq/applications.py
```diff
@@ -0,0 +1,52 @@
+import json
+import os
+
+import cirq
+import requests
+
+import cirq_superstaq as css
```
vtomole, line 7: I'll always bring this up: I'm uneasy about abbreviating imports like this. I feel like your package has to be "popular" enough to warrant this. For example, people know that `np` is numpy. They don't need to find where the import is stated to find this out.

cirq_superstaq/job_test.py
```diff
@@ -14,9 +14,10 @@
 
 from unittest import mock
 
-import cirq_superstaq as superstaq
 import pytest
 
+import cirq_superstaq as superstaq
```
vtomole, line 19: Let's revert this to `import cirq_superstaq` to discourage doing imports like this. In a situation where I import `cirq_superstaq` and `qiksit_superstaq` in the same file, which one deserves the `as superstaq` abbreviation?

cirq_superstaq/superstaq_client_test.py
```diff
@@ -15,10 +15,11 @@
 import json
 from unittest import mock
 
-import cirq_superstaq as superstaq
 import pytest
 import requests
 
+import cirq_superstaq as superstaq
```
vtomole, line 21: Revert to `import cirq_superstaq`

cirq_superstaq/applications_test.py
```diff
@@ -0,0 +1,47 @@
...
+from typing import Any
+from unittest.mock import patch
+
+import cirq
+import pytest
+
+import cirq_superstaq as css
+
+
+def test__get_api_url() -> None:
```
vtomole, line 11:
```suggestion
def test_get_api_url() -> None:
```

cirq_superstaq/applications_test.py
```diff
@@ -0,0 +1,47 @@
...
+    assert css.applications._get_api_url() == css.API_URL
+
+    with patch.dict(os.environ, {"SUPERSTAQ_REMOTE_HOST": "https://127.0.0.1:5000"}):
+        assert css.applications._get_api_url() == "https://127.0.0.1:5000"
+
+
+def test__get_headers() -> None:
```
vtomole, line 18:
```suggestion
def test_get_headers() -> None:
```

cirq_superstaq/applications_test.py
```diff
@@ -0,0 +1,47 @@
...
+        css.applications._get_headers()
+
+    with patch.dict(os.environ, {"SUPERSTAQ_API_KEY": "foobar"}):
+        assert css.applications._get_headers() == {
+            "Authorization": "foobar",
+            "Content-Type": "application/json",
+        }
+
+
+def test__should_verify_requests() -> None:
```
vtomole, line 29:
```suggestion
def test_should_verify_requests() -> None:
```

cirq_superstaq/applications.py
```diff
@@ -0,0 +1,52 @@
+import json
```
vtomole, line 1: The code in this file would be more appropriate in `superstaq_client.py` instead of `applications.py`.

vtomole (changes requested):

cirq_superstaq/applications.py
singular-value: Oh yeah, I meant to revert this. I'm going to do `css` in shell, but I agree that to look more like typical cirq code, we should do `from cirq_superstaq import blah` instead

cirq_superstaq/job_test.py
singular-value: Sure.
(To be clear, this was in the previous code-->I was just re-ordering to comply with my change to .flake8 where cirq_superstaq is now considered a local module)

cirq_superstaq/applications.py
singular-value: I agree medium-term, when we have a `/compile` endpoint as opposed to an aqt-specific compilation endpoint.

I see applications.py being fairly specific use cases, whereas superstaq_client.py being "cirq-general" use cases.

vtomole: How about renaming to `aqt.py`? That way we don't have finance, logistic, and finance stuff in the same file.

singular-value: Makes sense.
cc @pnoell, you'll want to put your stuff in finance.py and logistics.py --> I'm going to get rid of applications.py

pnoell: Sounds good, should I copy over the files? Or is there a way to mirror them from the SuperstaQ repo?

singular-value: @vtomole thoughts on where you think `_get_api_url()`, `_get_headers()`, and `_should_verify_requests()` should go?

I realize I put aqt_compile into applications.py because I wanted to have those three functions available throughout superstaq for any applications.


One idea is to move them into _init_vars.py

singular-value: > Sounds good, should I copy over the files? Or is there a way to mirror them from the SuperstaQ repo?

@pnoell so the logistics.py and finance.py files can stay in SuperstaQ, but in this repository, you can make two new files called logistics.py and finance.py-->those should be very "thin" files, basically just things like

```
def tsp(locs):
    code for making a POST request to the /tsp endpoint
```

vtomole: Before we discus that, this is how i'm seeing people using `aqt_compile`

```python
import cirq
import cirq_superstaq


q0 = cirq.LineQubit(0)
q1 = cirq.LineQubit(1)

circuit = cirq.Circuit(cirq.H(q0), cirq.CNOT(q0, q1), cirq.measure(q0))


service = cirq_superstaq.Service(
    api_key="""Insert superstaq token that you received from https://superstaq.super.tech""",
    verbose=True,
)

aqt_compiled_circuit = service.aqt_compile(circuit=circuit)
```

Does this look correct to you?

singular-value: Almost, except I don't see `service` as providing a specific `aqt_compile()` method but rather a general `compile` method:

```py
import cirq
import cirq_superstaq

q0 = cirq.LineQubit(0)
q1 = cirq.LineQubit(1)

circuit = cirq.Circuit(cirq.H(q0), cirq.CNOT(q0, q1), cirq.measure(q0))

service = cirq_superstaq.Service(
    api_key="""Insert superstaq token that you received from https://superstaq.super.tech""",
    verbose=True,
)

aqt_compiled_circuit = service.compile(circuit=circuit, target="aqt")
honeywell_compiled_circuit = service.compile(circuit=circuit, target="honeywell")
```

But we're not quite ready for providing that API yet-->hence why I was viewing `aqt_compile` as a specific application, not so far unlike finance or logistics.

vtomole: Let's add a `service.compile` that throws an error when the target that is not `aqt`

cirq_superstaq/superstaq_client_test.py
singular-value: resolved

cirq_superstaq/job_test.py
singular-value: resolved

cirq_superstaq/applications_test.py
singular-value: resolved

cirq_superstaq/applications_test.py
singular-value: no longer relevant

cirq_superstaq/applications.py
singular-value: discussed in person

setup.py
```diff
@@ -25,7 +25,11 @@
     "cirq_superstaq." + package for package in find_packages(where="cirq_superstaq")
 ]
 
-print(cirq_superstaq_packages)
+
+# Optional installations. For example, pip install cirq-superstaq[aqt] will include
```
vtomole, line 29: When I suggested this I was more thinking along the lines of having the `aqt.py` as a module that wouldn't be installed by default; but I don't think it's worth the amount of labor it's going to take messing around with Python's import system. 

Thoughts?

vtomole (changes requested):

cirq_superstaq/job_test.py
vtomole: 👍

cirq_superstaq/aqt.py
```diff
@@ -0,0 +1,29 @@
...
+    We only deserialize the qtrl sequence if qtrl is available locally.
+    """
+    out_circuit = cirq.read_json(json_text=json_dict["compiled_circuit"])
+
+    try:
+        import qtrl
+    except ModuleNotFoundError:
+        return out_circuit
+
+    if True:  # pragma: no cover b/c qtrl is not yet open source
```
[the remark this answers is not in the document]
singular-value: so that only one `# pragma: no cover` is needed

cirq_superstaq/aqt.py
```diff
@@ -0,0 +1,29 @@
...
+        state_str = json_dict["state_jp"]
+        state = pickle.loads(codecs.decode(state_str.encode(), "base64"))
+        seq = qtrl.sequencer.Sequence(n_elements=1)
+        seq.set_state(state)
+        seq.compile()
+        return (out_circuit, seq)
```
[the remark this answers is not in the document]
singular-value: 👍

cirq_superstaq/aqt.py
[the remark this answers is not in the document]
vtomole: 👍

cirq_superstaq/aqt.py
```diff
@@ -0,0 +1,29 @@
+import codecs
+import pickle
+from typing import Union
+
+import cirq
+
+
+def read_json(json_dict: dict) -> Union[cirq.Circuit, tuple]:
+    """Reads out returned JSON from SuperstaQ API's AQT compilation endpoint.
```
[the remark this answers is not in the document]
singular-value: Done

setup.py
singular-value: Agreed. Went back to original setup.py (minus a print statement)

cirq_superstaq/aqt.py
```diff
@@ -0,0 +1,42 @@
...
+import pickle
+from dataclasses import dataclass
+from typing import Optional
+
+import cirq
+
+try:
+    import qtrl
+except ModuleNotFoundError:
+    pass
```
singular-value, line 11: (this is needed for the type in L17)

cirq_superstaq/aqt.py
```diff
@@ -1,29 +1,42 @@
 import codecs
 import pickle
-from typing import Union
+from dataclasses import dataclass
+from typing import Optional
 
 import cirq
 
+try:
```
vtomole, line 8: Is this added just for the future or is it to make mypy happy?

singular-value: it's to make mypy happy

cirq_superstaq/aqt.py
vtomole: 👍

cirq_superstaq/aqt.py
vtomole: 👍

vtomole (approved):
LGTM

PR #9, opened 2021-08-02
Include finance and logistics endpoints in cirq-superstaq

pnoell:


singular-value (commented):
@vtomole where do you think
`_get_api_url()` and `_get_headers()` and `_should_verify_requests()` should belong?

vtomole:
@singular-value Let's move them to `superstaq_client.py`

vtomole (changes requested):
Call this code in `service.py` so that we may do something like: https://github.com/SupertechLabs/cirq-superstaq/pull/8#discussion_r681207907. Happy to discuss this in person.

singular-value:
Yup yup. I'm going to do my aqt changes (incorporating with service.py)
first and Peter will follow the format.

singular-value:
(made my changes in #8, @pnoell your changes will probably look similar. Basically the "entry point" to tsp/logistics/minvol/etc should be `Service`)

pnoell:
Created a separate PR. Closing this one as it's no longer relevant.

PR #11, opened 2021-08-07
Support use of "job_id" instead of "id"

singular-value:
SuperstaQ API now returns a "job_id" key. These changes are needed to comply with the new API.


vtomole (approved):
LGTM

PR #12, opened 2021-08-09
Release `version 0.0.1` to pypi

vtomole:
This will make it so users can perform `pip install cirq-superstaq`: https://pypi.org/project/cirq-superstaq/0.0.1/


singular-value (approved):
lgtm

Issue #14, opened 2021-08-10
Use __setstate__ for qtrl


singular-value:

singular-value:
I think this has been taken care of.

PR #21, opened 2021-08-11
Support for `qtrl.sequencer.__setstate__` (#14)

richrines1:


vtomole (approved):
This is fine to me, but note that it will break our version of qtrl unfortunately.  Please add a `__setstate__()` method here: https://github.com/SupertechLabs/qtrl/blob/5f1a19f968eea746b2501050bfe756bf4e51f954/qtrl/sequencer/sequencer.py#L998 that simply calls `self.set_state()`

vtomole:
@richrines1 Cool. Feel free to merge when you are ready.

PR #22, opened 2021-08-13
add import of aqt submodule

richrines1:
previously `$ check/pytest` and `$ pytest aqt_test.py service_test.py` both worked because aqt_test.py imports `cirq_superstaq.aqt`, but `$ pytest service_test.py` would fail because `cirq_superstaq.aqt.AQTCompilerOutput` was used before the `aqt` submodule was loaded


vtomole (approved):
LGTM

singular-value:
Sweet thanks.

Issue #16, opened 2021-08-10
Return pulse_list in AQT's CompilerOutput


singular-value:

singular-value:
Was closed by #46

PR #26, opened 2021-08-17
Include pulse list in AQTCompilerOutput

richrines1:
fixes: #16


cirq_superstaq/aqt.py
```diff
@@ -11,11 +11,18 @@
 except ModuleNotFoundError:
     pass
 
+PulseType = Union[
```
singular-value, line 14: commenting for url to link to

singular-value:
I'll close this PR since it was subsumed by #46

PR #27, opened 2021-08-17
Return `response.text` as the error message when a `400` status code is returned

vtomole:


cirq_superstaq/superstaq_client.py
```diff
@@ -257,6 +257,29 @@ def _target(self, target: Optional[str]) -> str:
...
+            raise cirq_superstaq.superstaq_exceptions.SuperstaQNotFoundException(
+                "SuperstaQ could not find requested resource."
+            )
+
+        if response.status_code not in self.RETRIABLE_STATUS_CODES:
+            message = response.reason
+            if response.status_code == 400:
+                message = str(response.text)
+            raise cirq_superstaq.superstaq_exceptions.SuperstaQException(
+                "Non-retry-able error making request to SuperstaQ API. "
```
richrines1, line 277: nit: I think this should be either `"Non-retriable"` or `"Non-retryable"`

richrines1 (approved):
lgtm

singular-value (approved):
👍

PR #30, opened 2021-08-23
`verify_https` boolean variable

vtomole:
The `verify` check works. We were comparing the wrong values `super.tech != super.tech/v0.1`


singular-value (approved):
cool

vtomole:
This problem doesn't exist in `qiksit-superstaq`.

PR #40, opened 2021-08-27
Support multi-element qtrl sequences

singular-value:


cirq_superstaq/aqt.py
```diff
@@ -14,34 +13,54 @@
...
+        circuits: Union[cirq.Circuit, List[cirq.Circuit]],
+        seq: Optional["qtrl.sequencer.Sequence"] = None,
+    ) -> None:
+        if isinstance(circuits, cirq.Circuit):
+            self.circuit = circuits
+        else:
+            self.circuits = circuits
+        self.seq = seq
 
+    def __repr__(self) -> str:
```
vtomole, line 28: Let's do

```python
  def __repr__(self) -> str:
        if hasattr(self, "circuit"):
            return f"AQTCompilerOutput({self.circuit!r}, {self.seq})"
        return f"AQTCompilerOutput({self.circuits!r}, {self.seq})"
```

vtomole (changes requested):
Small suggestion then LGTM

cirq_superstaq/aqt.py
singular-value: done

vtomole (approved):
LGTM

PR #43, opened 2021-09-01
Update cirq version to latest

vtomole:


requirements.txt
```diff
@@ -1,3 +1,3 @@
 applications-superstaq==0.0.3
-cirq
+cirq==0.13.0.dev20210901195302
```
richrines1, line 2: does cirq_superstaq require this exact version or would something like `cirq>=0.11` be okay here?

vtomole: Good question. I have to make it this version because anything with a "dev" in it (as you see here) won't be picked up automatically by pip. If i did do `cirq>=0.11` the cirq version that will be installed will be `0.12.0`.

vtomole: You are right. Just changed. Thanks!

richrines1 (approved):

cirq_superstaq/custom_gates.py
```diff
@@ -42,6 +40,9 @@ def __init__(self, theta: float) -> None:
         """
         self.theta = cirq.ops.fsim_gate._canonicalize(theta)  # between -pi and +pi
 
+    def _num_qubits_(self) -> int:
```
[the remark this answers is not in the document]
richrines1: I think `_num_qubits_()` gets called implicitly by `Gate.on()`/`Gate.__call__()`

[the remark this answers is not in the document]
vtomole: Might be wrong on this; haven't tested: This class now inherits from `cirq.Gate` which has a `def validate_args` that calls this method.

PR #46, opened 2021-09-02
Include pulse_list(s) in AQTCompilerOutput

### LINKED ISSUE #45: Include pulse_list(s) in AQTCompilerOutput

singular-value:
Closes #45

### DIFF
--- a/cirq_superstaq/aqt.py
+++ b/cirq_superstaq/aqt.py
@@ -18,17 +18,21 @@ class AQTCompilerOutput:
         self,
         circuits: Union[cirq.Circuit, List[cirq.Circuit]],
         seq: Optional["qtrl.sequencer.Sequence"] = None,
+        pulse_lists: Optional[Union[List[List], List[List[List]]]] = None,
     ) -> None:
         if isinstance(circuits, cirq.Circuit):
             self.circuit = circuits
+            self.pulse_list = pulse_lists
         else:
             self.circuits = circuits
+            self.pulse_lists = pulse_lists
+
         self.seq = seq
 
     def __repr__(self) -> str:
         if hasattr(self, "circuit"):
-            return f"AQTCompilerOutput({self.circuit!r}, {self.seq!r})"
-        return f"AQTCompilerOutput({self.circuits!r}, {self.seq!r})"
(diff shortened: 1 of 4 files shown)


cirq_superstaq/aqt.py
```diff
@@ -18,17 +18,21 @@ def __init__(
         self,
         circuits: Union[cirq.Circuit, List[cirq.Circuit]],
         seq: Optional["qtrl.sequencer.Sequence"] = None,
+        pulse_lists: Optional[Union[List[List], List[List[List]]]] = None,
```
vtomole, line 21: Just wondering: We don't return pulse lists every time?

vtomole (approved):
LGTM. For another PR: integration test for this.

richrines1 (approved):
lgtm, but out of curiosity - is there a reason not to define separate `AQTCompilerOutput` and `AQTCompilerOutputs` dataclasses with static attribute names rather than changing attributes based on the input type? it seems like it would clarify some logic and simplify testing a bit, and it's possibly non-ideal for downstream/user-created functions which take in a `AQTCompilerOutput` object to always need to do something like `if hasattr(obj, "circuits"): ...` to avoid throwing an attribute error

vtomole:
I prefer this because we have one `CompilerOutput` type. We can add getters to `AQTCompilerOutput` that return `None` instead of failing on the `hasattr` calls.

singular-value:
@vtomole the pulse list is a list of list of qtrl sequences. Since we don't have qtrl (not open source) on github, we won't be able to do a sensible integration check for it. Ditto for why we don't always return pulse_list(s). If qtrl is not installed in the user's environment, then we set pulse_list(s) to None.

singular-value:
@richrines1 I slightly favor the current approach here but definitely see where you're coming from.
For #agility let's stick with this for now, but we can revisit if/when there is a compelling reason to switch to having two classes/types.

vtomole:
> Since we don't have qtrl (not open source) on github, we won't be able to do a sensible integration check for it
 
Ah, this is an edge case for where we could add an internal daily integration test.

richrines1:
@singular-value @vtomole sounds good, I can definitely see it both ways. In the meantime it might be nice to add a property or method to check instead of having to use `hasattr`

singular-value:
Addressed in #48

PR #47, opened 2021-09-02
Daily integration check

singular-value:


.github/workflows/daily-integration-check.yml
```diff
@@ -0,0 +1,28 @@
+name: Daily Integration Check
+
+on:
+  schedule:
+  - cron: "0 14 * * *"  # Daily at 14:00 UTC
+  workflow_dispatch:  # so that we can also run manually, e.g. for testing
```
vtomole, line 6: How will this be triggered?

singular-value: Not sure if I'm answering your question, but it is triggered automatically (daily 9am central standard time) by the cron syntax here
Same as for our GoodMorning bot.

singular-value: Maybe you're asking "where will the output go"?
I'm not sure yet. I think it will email us when it fails.

vtomole: Sure. I was asking about `workflow_dispatch:  # so that we can also run manually, e.g. for testing`. How will it work?

vtomole (commented):
One con of this approach is that there will be test data in prod. IMO, that's not _really_ a big deal (agility I guess?). I can see how a Software Engineer more senior than me would hate this though.

vtomole:
Googling `testing in prod` show a lot of articles that don't discourage this so i think we are fine.

.github/workflows/daily-integration-check.yml
singular-value: oh, it's through https://github.com/SupertechLabs/cirq-superstaq/actions
Having the workflow_dispatch option enables manual trigger https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

vtomole: 👍

vtomole (approved):
LGTM

richrines1 (approved):
lgtm

PR #50, opened 2021-09-03
Interaction with `/ibm compile` endpoint and bump version to `0.0.8`

singular-value:
image


cirq_superstaq/service.py
```diff
@@ -209,6 +210,15 @@ def aqt_compile(
...
         return aqt.read_json(json_dict, circuits_list)
 
+    def ibm_compile(self, circuit: cirq.Circuit) -> Any:
+        """Returns pulse schedule for the given circuit, on IBM Jakarta 7Q device.
+
+        Qiskit must be installed for returned object to correctly deserialize to a pulse schedule.
+        """
+        serialized_program = cirq.to_json([circuit])
+        json_dict = self._client.ibm_compile(serialized_program)
+        return applications_superstaq.converters.deserialize(json_dict["pulses"])[0]
```
singular-value, line 220: I didn't end up making a new file and datastructure (akin to aqt.py and AQTCompilerOutput) because for now we're only returning the pulse schedule.

If qiskit is installed (with the right version), it works. If not, the deserialization will break with a sensible error: `ModuleNotFoundError: No module named 'qiskit.pulse'`

richrines1: would it make sense to instead throw a version of `cirq_superstaq.SuperstaQException` instead? that way it would be clear that it's intended behavior and not a bug

singular-value:
would it catch the ModuleNotFoundError? Would you mind adding it to this
commit?

richrines1:
@singular-value yeah I'd just set it up to catch `ModuleNotFoundError`s and throw something else - will add it now

cirq_superstaq/superstaq_exceptions.py
```diff
@@ -27,6 +27,16 @@ def __init__(self, message: str, status_code: int = None):
...
 
 
+class SuperstaQModuleNotFoundException(SuperstaQException):
+    """
+    An exception for SuperstaQ features requiring an uninstalled module."""
+
+    def __init__(self, name: str, context: str):
+        message = f"'{context}' requires module '{name}'"
+        super().__init__(message)
+        self.name = name
```
vtomole, line 37: Why are we setting this variable?

vtomole (changes requested):
Where is this sympy dependency coming from?

cirq_superstaq/superstaq_exceptions.py
richrines1: I just set it because it gets set in the `ModuleNotFoundError` builtin, but thinking about it more it shouldn't be. will remove

richrines1:
> Where is this sympy dependency coming from?

I'm not sure why it's used in `test_service_run()` - it seems like it could be removed. in `test_service_ibmq_compile()` it was just used to make pickle throw a `ModuleNotFoundError` (just removed)

vtomole (approved):

vtomole:
This PR was also used to deploy the newest version of `cirq-superstaq`

richrines1:
> I'm not sure why it's used in `test_service_run()` - it seems like it could be removed. in `test_service_ibmq_compile()` it was just used to make pickle throw a `ModuleNotFoundError` (just removed)

actually sympy is necessary for `test_service_run()` to test the `param_resolver` functionality in `service.run` so shouldn't be removed. otherwise lgtm

richrines1 (approved):

Issue #63, opened 2021-10-03
Daily integration test is failing


singular-value:
https://github.com/SupertechLabs/cirq-superstaq/actions/runs/1300641118

Even after fixing server

singular-value:
Was fixed via #64 and via a server-side change

vtomole:
Failing again for different reasons will re-open instead of opening a new issue.

PR #64, opened 2021-10-03
ask for float instead of str in get_balance

singular-value:
Related* to #63 but actually the real fix is server-side

*in the sense that this fix is also needed to make the daily integration test for balance to actually work


vtomole (approved):
LGTM

PR #65, opened 2021-10-05
adds pypi publish action

dowusu-antwi:
need to create and include github secret in ${{ secrets.PYPI_API_TOKEN }}

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,0 +1,7 @@
+# This publishes to PyPI.
+- name: Publish package
+  if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+  uses: pypa/gh-action-pypi-publish@release/v1
+  with:
+    user: __token__
+    password: ${{ secrets.PYPI_API_TOKEN }}


dowusu-antwi:
Edited `main.yaml`, for syntax errors and to update `_version.py` on every push with a version-tagged commit.

dowusu-antwi (commented):
@vtomole it's possible writing a shell script is simpler? It seems simpler to me to include the commands directly rather than figuring out how to get the yaml file to run an additional shell script.

@richrines1, I suppose it depends on whether you add a tag to commits to `circ-superstaq` as per https://stackoverflow.com/questions/29830544/how-do-i-add-a-tag-to-a-github-commit, so we may want to try that.

dowusu-antwi:
Shell script has a syntax error -- working on that now...

vtomole:
Testing with the aynch collector seems cost us.

`check/all` with with this PR

```
real	0m43.599s
user	0m35.571s
sys	0m1.065s
```

Default `check/all`

```
The HTML pages are in build/html.

real	0m31.877s
user	0m25.612s
sys	0m0.885s

```

dowusu-antwi (commented):
@richrines1 to reiterate this assumes that just before pushing, commits are tagged as seen here: https://stackoverflow.com/questions/29830544/how-do-i-add-a-tag-to-a-github-commit

(also assuming we are tagging commits in general, which @vtomole mentions in https://github.com/SupertechLabs/SuperstaQ/issues/838 is new for SuperstaQ)

vtomole (approved):
We need to merge this in to actually test.

PR #81, opened 2021-10-07
Fix service import

dowusu-antwi:
Added Service object import to fix undefined object error.

### DIFF
--- a/examples/ibmq_compile.ipynb
+++ b/examples/ibmq_compile.ipynb
@@ -8,7 +8,9 @@
    "outputs": [],
    "source": [
     "import cirq\n",
-    "import cirq_superstaq"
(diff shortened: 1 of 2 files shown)


examples/ibmq_compile.ipynb
```diff
@@ -8,7 +8,9 @@
    "outputs": [],
    "source": [
     "import cirq\n",
-    "import cirq_superstaq"
+    "import cirq_superstaq\n",
```
singular-value, line 11: Can you do `from cirq_superstaq import AceCRPlusMinus, Service` on the line below, and then we don't even need `import cirq_superstaq`

singular-value (approved):
lgtm

PR #89, opened 2021-10-13
Deserialize circuits to always return a list

richrines1:
this makes it symmetric with `qiskit_superstaq.serialization.deserialize_circuits()` (which uses `qpy_serialization.load()` which always returns a list of circuits). It also simplifies type checking a bit

### DIFF
--- a/cirq_superstaq/aqt.py
+++ b/cirq_superstaq/aqt.py
@@ -68,9 +68,8 @@ def read_json(json_dict: dict, circuits_list: bool) -> AQTCompilerOutput:
 
         pulse_lists = applications_superstaq.converters.deserialize(json_dict["pulse_lists_jp"])
 
-    resolvers = [cirq_superstaq.custom_gates.custom_resolver, *cirq.DEFAULT_RESOLVERS]
     compiled_circuits = [
-        cirq.read_json(json_text=c, resolvers=resolvers) for c in json_dict["cirq_circuits"]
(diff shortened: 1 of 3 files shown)


cirq_superstaq/aqt.py
```diff
@@ -68,9 +68,8 @@ def read_json(json_dict: dict, circuits_list: bool) -> AQTCompilerOutput:
 
         pulse_lists = applications_superstaq.converters.deserialize(json_dict["pulse_lists_jp"])
 
-    resolvers = [cirq_superstaq.custom_gates.custom_resolver, *cirq.DEFAULT_RESOLVERS]
     compiled_circuits = [
-        cirq.read_json(json_text=c, resolvers=resolvers) for c in json_dict["cirq_circuits"]
+        cirq_superstaq.serialization.deserialize_circuits(c)[0] for c in json_dict["cirq_circuits"]
```
vtomole, line 72: What happens when we do

```python
compiled_circuits = cirq_superstaq.serialization.deserialize_circuits(json_dict["cirq_circuits"])
```

vtomole: 👍

vtomole (changes requested):
Let's put all the client changes related to the API on hold until we update the API.

richrines1:
this was absorbed into #83. closing.

PR #99, opened 2021-10-26
Support AceCR with an Rx() gate on the target qubit

singular-value:
This enables new optimizations that our Argonne users want


dowusu-antwi:
Merging `main` into this branch to get updates from PR #100 should fix the error:

cirq-version-error

singular-value:
Tracking internally

vtomole:
The problem wasn't the Cirq version, the `cirq-superstaq` code didn't call off to `measurement_key_names`.

PR #111, opened 2021-11-02
fixed test to check for all backends regardless of ordering

### LINKED ISSUE #63: Daily integration test is failing

dowusu-antwi:
Changed test to check for all backend strings (ordering is irrelevant). Fixes #63.

daily-integration-success

### DIFF
--- a/cirq_superstaq/daily_integration_test.py
+++ b/cirq_superstaq/daily_integration_test.py
@@ -65,9 +65,14 @@ def test_get_backends(service: cirq_superstaq.Service) -> None:
             "d-wave_advantage-system1.1_qpu",
             "ionq_ion_qpu",
         ],
-        "compile-only": ["aqt_keysight_qpu", "sandia_qscout_qpu"],
+        "compile-only": [
+            "aqt_keysight_qpu",
+            "sandia_qscout_qpu",
+        ],
     }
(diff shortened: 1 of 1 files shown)


richrines1 (dismissed):
lgtm

cirq_superstaq/daily_integration_test.py
```diff
@@ -67,7 +67,9 @@ def test_get_backends(service: cirq_superstaq.Service) -> None:
         ],
         "compile-only": ["aqt_keysight_qpu", "sandia_qscout_qpu"],
     }
-    assert service.get_backends() == expected
+    result = service.get_backends()
+    assert set(result["compile-and-run"]) == set(expected["compile-and-run"])
```
vtomole, line 71: Why call `set` instead of `sort`?

vtomole (changes requested):

cirq_superstaq/daily_integration_test.py
dowusu-antwi: No particular reason -- sorting makes sense though. Done.

vtomole (approved):
LGTM

PR #115, opened 2021-11-09
Updates `finance` notebook for new deployment

dowusu-antwi:
- changed IBM Q device to `jakarta`
- removed the retired / offline `rigetti` device
- fixed backend names to fit new format
finance-fix-00
finance-update-01


vtomole (approved):
LGTM

PR #118, opened 2021-11-10
Pass in `qubo` result directly

dowusu-antwi:
(+ nit: convert docstring to string)

qubo-solved


vtomole:
Will defer to @singular-value

singular-value (approved):
looks good thanks

PR #121, opened 2021-11-11
Updated compile-only devices

dowusu-antwi:
To match deployed version of `GetBackends` endpoint.

As verbalized by @vtomole, the test is indeed showing us the kind of errors we want it to (devices are not taken offline as often as was feared, which would cause the test to fail each time).

Somewhat tangentially related: perhaps it's a good idea to include in the docs standard requirements for deployment? (If that's something we can standardize.)

### DIFF
--- a/cirq_superstaq/daily_integration_test.py
+++ b/cirq_superstaq/daily_integration_test.py
@@ -63,7 +63,12 @@ def test_get_backends(service: cirq_superstaq.Service) -> None:
             "aws_tn1_simulator",
             "ionq_ion_qpu",
         ],
-        "compile-only": ["aqt_keysight_qpu", "sandia_qscout_qpu"],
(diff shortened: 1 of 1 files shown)


vtomole:
> standard requirements for deployment

Requirements like what?

vtomole (approved):
LGTM

dowusu-antwi:
> Requirements like what?

Something along the lines of "you should deploy when <this much change has been made>". If it's not something we can measure, then it could be useful to at least document how regular or semi-regular we want deployments to be.

vtomole:
For agility reasons, let's just keep it to "Don't deploy if the `daily_integration_test` is not passing" for now.

vtomole:
Maybe there is a way for us to script it where it fails to deploy if the integration test has failed.

PR #123, opened 2021-11-15
Inherit `Logistics` and `Finance` from `applications-superstaq`

vtomole:

PR #124, opened 2021-11-16
Add `__repr__` for `class Job`

vtomole:
Fixes: https://github.com/SupertechLabs/cirq-superstaq/issues/97

### DIFF
--- a/cirq_superstaq/job.py
+++ b/cirq_superstaq/job.py
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Represents a job created via the SuperstaQ API."""
-
 import collections
 import time
+from typing import Any
 
 import applications_superstaq
 from applications_superstaq import superstaq_client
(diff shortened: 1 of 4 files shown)


cirq_superstaq/service.py
```diff
@@ -77,7 +76,7 @@ def counts_to_results(
     return result
 
 
-class Service:
+class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
```
richrines1, line 79: is this from #123?

vtomole: Yes. I've merged `main` into this branch to get rid of that diff.

cirq_superstaq/job.py
```diff
@@ -169,3 +169,11 @@ def counts(
...
     def __str__(self) -> str:
         return f"Job with job_id={self.job_id()}"
+
+    def __repr__(self) -> str:
+        return f"cirq_superstaq.Job(client={self._client!r}, job_dict={self._job!r})"
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(self, type(other)):
+            return False
+        return self._job == other._job
```
richrines1, line 179: is there any reason we'd want to check `self._client == other._client` as well?

vtomole: I think this is good enough of a comparison because the backend should never duplicate `job_ids`

richrines1 (approved):
lgtm

PR #127, opened 2021-11-18
Removed devices not-accessible to non-Super.tech users

dowusu-antwi:
Integrating `pulser` neutral atom device.

(Integration test coming soon...)


vtomole (dismissed):
LGTM

PR #129, opened 2021-11-18
Checking devices accessible to Super.tech users in `daily_integration`

dowusu-antwi:
Matching backends returned for test user to `daily_integration` test.

Moved https://github.com/SupertechLabs/cirq-superstaq/pull/127 to this -- error turned out not to be a problem with `neutral_atom_qpu`, as those changes (https://github.com/SupertechLabs/SuperstaQ/pull/917#issuecomment-972362675) are not yet merged.

backend-daily-success


dowusu-antwi:
(from verbalized discussion with @vtomole) For now it seems best to check if a couple of devices we know we want to have are returned for the daily tests.

vtomole (approved):
LGTM

PR #133, opened 2021-12-06
Added check for an `ibmq_token`

### LINKED ISSUE #119: Make token required for IBM Q devices
For IBM Q device requests, check provider parameters (i.e., where provider refers to `(group, project, hub)`) and throw exception if not set.

dowusu-antwi:
Requires an `ibmq_token` for IBM Q targets. Fixes #119.

This only checks for the token in `create_job`, making token parameter itself in `Service` optional so that clients can run on non-IBM Q devices without needing an IBM Q token.

### DIFF
--- a/cirq_superstaq/service.py
+++ b/cirq_superstaq/service.py
@@ -220,12 +220,11 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         Raises:
             SuperstaQException: If there was an error accessing the API.
         """
-        if target:
-            if "ibmq" in target and not self.ibmq_token:
-                raise EnvironmentError(
-                    "Parameter ibmq_token was not specified and the environment variable IBMQ_TOKEN"
-                    "was also not set."
-                )
+        if target and "ibmq" in target and not self.ibmq_token:
+            raise EnvironmentError(
+                "Parameter ibmq_token was not specified and the environment variable IBMQ_TOKEN was"
+                "also not set."
+            )
         serialized_circuits = cirq_superstaq.serialization.serialize_circuits(circuit)
         result = self._client.create_job(
             serialized_circuits={"cirq_circuits": serialized_circuits},
--- a/cirq_superstaq/service.py
+++ b/cirq_superstaq/service.py
@@ -89,6 +89,7 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         self,
         remote_host: Optional[str] = None,
         api_key: Optional[str] = None,
+        ibmq_token: Optional[str] = None,
         default_target: str = None,
         api_version: str = cirq_superstaq.API_VERSION,
         max_retry_seconds: int = 3600,
@@ -126,6 +127,7 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
                 "Parameter api_key was not specified and the environment variable "
                 "SUPERSTAQ_API_KEY was also not set."
             )
+        self.ibmq_token = ibmq_token or os.getenv("IBMQ_TOKEN")
         self._client = superstaq_client._SuperstaQClient(
             client_name="cirq-superstaq",
             remote_host=self.remote_host,
@@ -218,6 +220,12 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         Raises:
             SuperstaQException: If there was an error accessing the API.
         """
+        if target:
+            if "ibmq" in target and not self.ibmq_token:
+                raise EnvironmentError(
+                    "Parameter ibmq_token was not specified and the environment variable IBMQ_TOKEN"
+                    "was also not set."
+                )
         serialized_circuits = cirq_superstaq.serialization.serialize_circuits(circuit)
         result = self._client.create_job(
             serialized_circuits={"cirq_circuits": serialized_circuits},
--- a/cirq_superstaq/service_test.py
+++ b/cirq_superstaq/service_test.py
@@ -52,7 +52,9 @@ def test_counts_to_results() -> None:
 
 
 def test_service_run_and_get_counts() -> None:
-    service = cirq_superstaq.Service(remote_host="http://example.com", api_key="key")
+    service = cirq_superstaq.Service(
+        remote_host="http://example.com", api_key="key", ibmq_token="token"
+    )
     mock_client = mock.MagicMock()
     mock_client.create_job.return_value = {
         "job_ids": ["job_id"],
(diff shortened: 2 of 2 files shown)


perlinm (dismissed):
:+1:

cirq_superstaq/service_test.py
```diff
@@ -161,6 +165,14 @@ def test_service_create_job() -> None:
     assert create_job_kwargs["target"] == "qpu"
 
 
+def test_service_create_job_no_ibmq_token() -> None:
+    service = cirq_superstaq.Service(remote_host="http://example.com", api_key="key")
+
+    circuit = cirq.Circuit(cirq.X(cirq.LineQubit(0)))
+    with pytest.raises(EnvironmentError):
+        _ = service.create_job(circuit=circuit, repetitions=100, target="ibmq_qpu")
```
richrines1, line 173: will this test fail if IBMQ_TOKEN is set as an environment variable?

cirq_superstaq/service.py
```diff
@@ -218,6 +220,12 @@ def create_job(
         Raises:
             SuperstaQException: If there was an error accessing the API.
         """
+        if target:
+            if "ibmq" in target and not self.ibmq_token:
```
perlinm, line 224: Asking as a novice developer: is there a reason not to have this all one one line:
`if target and "ibmq" in target and not self.ibmq_token:`

cirq_superstaq/service_test.py
perlinm: Hmm, it looks like environment variables have to be explicitly passed to `pytest` in order for it to pick them up.  See [example](https://stackoverflow.com/questions/36141024/how-to-pass-environment-variables-to-pytest) in which somebody *wants* to pass `pytest` environment variables, but does not know how.  So... I would guess `service` should not pick up on an IBMQ_TOKEN environment variable?

cirq_superstaq/service.py
dowusu-antwi: Good point. Fixed.

perlinm: Haha in that case my comment was quite petty.  Apologies :upside_down_face:

cirq_superstaq/service_test.py
dowusu-antwi: Yes, locally the test does end up failing if I do something like an `export IBMQ_TOKEN="foo"`.

Perhaps it's a good idea to set the environment variable to empty in the test? The idea is to test for the `EnvironmentError` in `create_job`:
https://github.com/SupertechLabs/cirq-superstaq/blob/11b869efcf095b78fae89efdbb5008f359a3812b/cirq_superstaq/service.py#L224-L227

dowusu-antwi: Update: with `os.environ["IBMQ_TOKEN"] = ""`, after setting `$IBMQ_TOKEN` manually still causes test to fail, but the error is something to the effect of a `requests` response error...looking into it now.

perlinm: I agree, simplest solution here would be to add something like `os.environ["IBMQ_TOKEN"] = None` right before making the service inside the test:
https://github.com/SupertechLabs/cirq-superstaq/blob/11b869efcf095b78fae89efdbb5008f359a3812b/cirq_superstaq/service_test.py#L168-L169
It feels a bit hackish though.  Ideally, we would want to isolate pytest from environment variables somehow...

Follow-up thought: we may want follow-up tests to access the environment variable, in which case we would need to do something like
```
old_token = os.gevenv("IBMQ_TOKEN")
os.environ["IBMQ_TOKEN"] = None
service = ...
os.environ["IBMQ_TOKEN"] = old_token
```
But that's *really* hackish.  Maybe it's okay because I can't imagine this being used anywhere other than a test for the absence of a token.

richrines1: you can use `monkeypatch` to clear the env variable for just the one check - cf. SupertechLabs/qiskit-superstaq#63

perlinm: Oh: would passing `token=""` to `service` work?

cirq_superstaq/service.py
dowusu-antwi: Not at all! Better-looking code is always a good idea. Also nice to take advantage of left->right evaluation. Thanks!

cirq_superstaq/service_test.py
dowusu-antwi: > Oh: would passing token="" to service work?

Yes -- right now it just checks to see if an IBM Q token is provided. To validate it, we would need to use `qiskit.IBMQ.save_account()` and `qiskit.IBMQ.load_account()`.

Actually, now that you mention it I guess it's better to include token validation in this PR as well. What gives me pause is then needing an `import qiskit` in `css`, which I'm not sure fits our design principles? (@richrines1)

richrines1: down the road another option might be to add a `conftest.py` file and use it to consistently set/clear environment variables across all the tests

richrines1: > Actually, now that you mention it I guess it's better to include token validation in this PR as well. What gives me pause is then needing an import qiskit in css, which I'm not sure fits our design principles? (@richrines1)

I think the checks should all pass without `qiskit`

richrines1: i'm not sure if validating the IBMQ_TOKEN is really necessary here. it's pretty reasonable to use a placeholder token for testing/development. the validity of the token only really matters if the user is submitting jobs that require it, at which point it should be handled by the API

dowusu-antwi: @perlinm, update: in fact, passing `token=""` will not work, since the environment variables are set to the empty string by default, rather than `None` (then taking advantage of `not ""` evaluating to `True` in the if condition). 

Although yes, I see what you're getting at, e.g., passing `token="foo"` to `service` will work. Per @richrines1's comment above, the idea is that we just want to make sure that a string is passed for the token, and the validation will be handled via SuperstaQ (although I don't think this is currently set up - there's an issue for it).

cirq_superstaq/service_test.py
```diff
@@ -161,6 +165,15 @@ def test_service_create_job() -> None:
     assert create_job_kwargs["target"] == "qpu"
 
 
+@mock.patch.dict(os.environ, {"IBMQ_TOKEN": ""})
```
richrines1, line 168: nice

cirq_superstaq/service_test.py
```diff
@@ -161,6 +165,15 @@ def test_service_create_job() -> None:
...
+def test_service_create_job_no_ibmq_token() -> None:
+    service = cirq_superstaq.Service(remote_host="http://example.com", api_key="key")
+
+    circuit = cirq.Circuit(cirq.X(cirq.LineQubit(0)))
+    with pytest.raises(EnvironmentError):
```
richrines1, line 173: nit
```suggestion
    with pytest.raises(EnvironmentError, match="Parameter ibmq_token was not specified"):
```

richrines1 (approved):
lgtm

PR #134, opened 2021-12-10
Updates `jupyter` notebook demo to include `ibmq_token` parameter in `Service`

dowusu-antwi:
To match changes in https://github.com/SupertechLabs/cirq-superstaq/pull/133, adds token parameter to jupyter notebook demo `Service` object.

(Also some clean-up, removed output warnings from `qscout_compile.ipynb`)

token-given

### DIFF
--- a/examples/finance.ipynb
+++ b/examples/finance.ipynb
@@ -17,7 +17,8 @@
    "source": [
     "from cirq_superstaq import Service\n",
     "service = Service(\n",
-    "    api_key=\"Insert superstaq token that you received from https://superstaq.super.tech\",\n",
+    "    api_key=\"Insert superstaq token that you received from https://superstaq.super.tech.\",\n",
+    "    ibmq_token=\"Insert IBM Q token that you received from https://quantum-computing.ibm.com/.\"\n",
     ")"
    ]
   },
@@ -25,7 +26,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "f13ac60d-5e5f-400c-bc03-6f52fd0242a9",
-   "metadata": {},
+   "metadata": {
+    "scrolled": true
+   },
    "outputs": [],
    "source": [
     "service.get_balance()"
@@ -213,7 +216,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.8.12"
+   "version": "3.8.10"
   }
  },
  "nbformat": 4,
--- a/examples/qscout_compile.ipynb
+++ b/examples/qscout_compile.ipynb
@@ -10,7 +10,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
    "id": "029917c7-091c-4563-98b2-b1bf4d532660",
    "metadata": {},
    "outputs": [],
@@ -21,7 +21,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 3,
    "id": "c80cdc6b",
    "metadata": {},
    "outputs": [],
@@ -44,7 +44,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 4,
    "id": "5f107c0d-426e-4444-9749-8589bea4b679",
    "metadata": {
     "scrolled": true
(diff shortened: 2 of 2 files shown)


singular-value:
I've lost some context here, so leaving review for @richrines1 / @vtomole 

One user experience question though: does this mean that the user always has to explicitly provide the IBM token? It would be nice it we can read from ~/.qiskit/qiskitrc by default.

vtomole:
Getting the token from `qiskitrc` and failing when it's not found when
trying to use the relevant APIs is the best way to go.

dowusu-antwi:
> One user experience question though: does this mean that the user always has to explicitly provide the IBM token? It would be nice it we can read from ~/.qiskit/qiskitrc by default.

@singular-value, the user need only provide their IBM token for creating / running jobs on IBM devices -- for non-IBM devices, no errors will be raised if they leave out the token parameter in `Service`.

Getting it from `qiskitrc` sounds like a good idea. One question though (also for @vtomole): does this mean that, rather than provide the token as a parameter, the user will need to load their account themselves (at least once, to load the value into `qiskitrc`)? I.e., inside the `css` notebook above, the user would do something like this:

```
IBMQ.save_account("token", overwrite=True)
if not IBMQ.active_account():
    IBMQ.load_account()
```

richrines1:
fwiw qiskit will also try to read a token from the `QE_TOKEN` environment variable, if it exists, and then fall back on `qiskitrc`

imo I think it makes sense to (1) give the user the option to specify `ibmq_token` as an optional argument in `Service()`, (2) if no token is passed attempt to read it from `os.getenv("QE_TOKEN")`, (3) finally attempt to read it from `~/.qiskit/qiskitrc`

singular-value:
> imo I think it makes sense to (1) give the user the option to specify ibmq_token as an optional argument in Service(), (2) if no token is passed attempt to read it from os.getenv("QE_TOKEN"), (3) finally attempt to read it from ~/.qiskit/qiskitrc

Yup, agree with this.

dowusu-antwi:
@richrines1, yeah that makes sense to me too. However, I'm thinking now that we only need to handle (1), since (2) and (3) are handled by `qiskit.IBMQ.save_account` and `qiskit.IBMQ.load_account`. I.e., `save_account` writes to `qiskitrc` and `load_acount` will read from `QE_TOKEN` (? not sure about that actually) and `qiskitrc`.

singular-value:
^ Yup, we don't want to have qiskit as a dependency for cirq-superstaq (and we don't want to have cirq as a dependency for qiskit-superstaq)

dowusu-antwi:
> I think we should avoid using `load_account()`/`save_account()` here - ideally the user shouldn't be required to have qiskit installed at all to use cirq-superstaq

Ah yes -- this work would be done on `SuperstaQ`, just as our other `qiskit`-related methods for, say, running on an IBM Q backend are technically "used" by `css`, but on the `SuperstaQ` side. (Perhaps this is not the best way to describe it.)

This might be a bigger design issue, but to clarify: right now `css` only checks to see if a token is provided at all, without validation. Validation is to be handled by `SuperstaQ`. From discussion it appears that: (1) on `css`, the necessary change is to make the `ibmq_token` parameter optional in `Service`, and (2) we should try to read the token in from either the optional `ibmq_token`, `QE_TOKEN`, or `qiskitrc` (in that order).

If this sounds right, I'll go ahead and try to make those changes on the `css`-side. However, since validation would be done on `SuperstaQ` anyway (where `qiskit` is accessible), I was thinking that once that's completed it might not be necessary to do (2), but perhaps it's better to pass on that for now.

richrines1:
another thought: it seems like this should maybe all be moved to applications-superstaq - possibly in [`user_config.UserConfig`](https://github.com/SupertechLabs/applications-superstaq/blob/main/applications_superstaq/user_config.py)? (which is already a parent class of `cirq_superstaq.Service`)

richrines1:
> the necessary change is to make the ibmq_token parameter optional in Service, and (2) we should try to read the token in from either the optional ibmq_token, QE_TOKEN, or qiskitrc (in that order).

(otherwise this sounds right to me)

singular-value:
I think this PR can now be abandoned. Instead, we are going to have a `.set_ibmq_token()` method

dowusu-antwi:
> I think this PR can now be abandoned. Instead, we are going to have a `.set_ibmq_token()` method

👍

PR #141, opened 2021-12-21
Rxy _qasm_ method

richrines1:
adds `_qasm_` to `custom_gates.Rxy`


cirq_superstaq/custom_gates_test.py
```diff
@@ -349,6 +352,21 @@ def test_rxy() -> None:
...
+        """\
+        OPENQASM 2.0;
+        include "qelib1.inc";
+
+
+        // Qubits: [0]
+        qreg q[1];
+
+
+        rxy(pi*-0.77,pi*4.56) q[0];
```
vtomole, line 365: Why is `phase_exponent` similar to `rot_angle`?  I think the gate we are using is

`  rot_gate = cirq_superstaq.custom_gates.Rxy(1.23 * np.pi, 4.56 * np.pi)`

vtomole (dismissed):
LGTM

vtomole (approved):
Re-approving.

cirq_superstaq/custom_gates_test.py
vtomole: Turns out Cirq does this.

PR #142, opened 2021-12-23
adding custom ParallelRxy gate

perlinm:
Moving ParallelRxy gate here from `superstaq.cq_device`


cirq_superstaq/custom_gates.py
```diff
@@ -434,6 +437,56 @@ def _from_json_dict_(cls, axis_angle: float, rot_angle: float, **kwargs: Any) ->
...
+    def exponent(self) -> float:
+        return cast(cirq_superstaq.Rxy, self.sub_gate).exponent
+
+    @property
+    def axis_angle(self) -> float:
+        return cast(cirq_superstaq.Rxy, self.sub_gate).axis_angle
+
+    @property
+    def rot_angle(self) -> float:
+        return cast(cirq_superstaq.Rxy, self.sub_gate).rot_angle
```
perlinm, line 464: This `cast`ing is definitely ugly.  I am open ears for ideas to work around it.  If you don't have the `cast`, then `mypy` complains about the following:
image

perlinm: To be honest, though, I don't think this matters, as `ParallelRxy` is hard-coded to have `Rxy` as a sub-gate.  Unless there is particularly slick wizardry to work around this, I vote to leave it as is.

cirq_superstaq/custom_gates_test.py
```diff
@@ -353,6 +356,40 @@ def test_rxy() -> None:
     cirq.testing.assert_has_diagram(circuit, "0: ───Rxy(π, 0.5π)───")
 
 
+def test_parallel_rxy() -> None:
+    qubits = cirq.LineQubit.range(2)
+
+    rot_gate = cirq_superstaq.custom_gates.ParallelRxy(len(qubits), 1.23 * np.pi, 4.56 * np.pi)
```
richrines1, line 362: nit
```suggestion
    rot_gate = cirq_superstaq.ParallelRxy(len(qubits), 1.23 * np.pi, 4.56 * np.pi)
```
(various instances above/below as well)

cirq_superstaq/custom_gates.py
```diff
@@ -434,6 +437,56 @@ def _from_json_dict_(cls, axis_angle: float, rot_angle: float, **kwargs: Any) ->
...
+
+    def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.CircuitDiagramInfo:
+        diagram_info = cirq.protocols.circuit_diagram_info(self.sub_gate, args, NotImplemented)
+        wire_symbols = tuple(diagram_info.wire_symbols) + ("·",) * (self.num_copies - 1)
+        return cirq.protocols.CircuitDiagramInfo(wire_symbols=wire_symbols)
+
+    def __str__(self) -> str:
+        return f"Rxy({self.axis_angle/np.pi}π, {self.rot_angle/np.pi}π) x {self.num_copies}"
```
richrines1, line 472: nit
```suggestion
        return f"Rxy({self.phase_exponent}π, {self.exponent}π) x {self.num_copies}"
```

cirq_superstaq/custom_gates.py
```diff
@@ -434,6 +437,56 @@ def _from_json_dict_(cls, axis_angle: float, rot_angle: float, **kwargs: Any) ->
         return cls(axis_angle, rot_angle)
 
 
+@cirq.value_equality(approximate=True)
+class ParallelRxy(cirq.ParallelGate):
+    """Wrapper class to define a ParallelGate of identical Rxy gates."""
+
+    def __init__(self, num_copies: int, axis_angle: float, rot_angle: float) -> None:
```
richrines1, line 444: defer to your judgement, but it's a little weird to me that num_copies is the first argument here but the last argument of `cirq.ParallelGate`

cirq_superstaq/custom_gates.py
richrines1: not sure if this is better but one way is to override `ParallelGate.sub_gate`:
```
    @property
    def sub_gate(self) -> cirq_superstaq.Rxy
        return self._sub_gate
```
and re-annotate `_sub_gate`:
```
    def __init__(self, num_copies: int, axis_angle: float, rot_angle: float) -> None:
        self._sub_gate: cirq.Rxy
        super().__init__(cirq_superstaq.Rxy(axis_angle, rot_angle), num_copies)
```

richrines1: actually much better: just re-annotate `_sub_gate` and use `self._sub_gate` instead of `sub_gate` below

cirq_superstaq/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.1.5"
+__version__ = "0.1.6"
```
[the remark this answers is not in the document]
perlinm: I think the point is to publish immediately.  I'm hoping to use this in `cq_device.py` right away.  Should this still be changed in a different PR?

cirq_superstaq/custom_gates.py
perlinm: Agreed; moving to last

cirq_superstaq/custom_gates.py
```diff
@@ -434,6 +437,58 @@ def _from_json_dict_(cls, axis_angle: float, rot_angle: float, **kwargs: Any) ->
...
+        self._phase_exponent = axis_angle / np.pi
+        self._exponent = rot_angle / np.pi
+        super().__init__(cirq_superstaq.Rxy(axis_angle, rot_angle), num_copies)
+
+    def on(self, *qubits: cirq.Qid) -> cirq.Operation:
```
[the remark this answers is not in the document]
perlinm: Yes, but then the qubit order is unpredictable/random... which doesn't matter functionally, but it's not so nice when printing a circuit with this gate (e.g. for the CQ demo).

cirq_superstaq/custom_gates.py
perlinm: I think I'll overload the `@property` as well, just so that it's accessible elsewhere:

```

    def __init__(self, axis_angle: float, rot_angle: float, num_copies: int) -> None:
        super().__init__(cirq_superstaq.Rxy(axis_angle, rot_angle), num_copies)
        self._sub_gate: Rxy

    @property
    def sub_gate(self) -> Rxy:
        return self._sub_gate
```

cirq_superstaq/custom_gates.py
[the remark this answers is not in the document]
richrines1: I guess I would avoid making things work differently from what a user would expect from `cirq.ParallelGate`/other cirq.Gates - e.g. one might reasonably expect the `op.qubits` to be in the same order that the qubits were passed. Could the qubits be sorted externally (i.e. prior to be passed to `.on()`) in cases in which it's nice to show them sorted?

cirq_superstaq/_version.py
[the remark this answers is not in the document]
richrines1: maybe doesn't matter but i think it's possibly good practice to make it it's own PR? There's always a small chance we'll want to sneak in something else between this being merged and publication. Also at least personally i've found it helpful in terms of keeping track of/searching for releases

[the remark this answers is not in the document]
perlinm: fair enough.  done.

cirq_superstaq/custom_gates.py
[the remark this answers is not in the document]
richrines1: somewhat related - we could also make `ParallelRxyGate` inherit from `cirq.InterchangeableQubitsGate` which would at least make it true that e.g. `ParallelRxyGate(x,y,3)(q0,q1,q2) == ParallelRxyGate(x,y,3)(q1,q2,q0)`/etc. (tbh i'm pretty surprised `cirq.ParallelGate` doesn't already do this)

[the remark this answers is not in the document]
perlinm: Surprisingly, `cirq.ParallelGate` [does not](https://github.com/quantumlib/Cirq/blob/v0.13.1/cirq-core/cirq/ops/parallel_gate.py#L30-L159).

Good idea for `ParallelRxy` though.  I'll add it.

richrines1 (approved):
cool, lgtm.

perlinm:
sweet thanks for the feedback / review :)

PR #144, opened 2021-12-23
Remove Cirq upper bound requirement 

perlinm:


vtomole (approved):

vtomole:
Ready for review.

richrines1:
I think we need to update the lower bound as well - e.g. 0.11 is broken due to `Circuit.all_measurement_key_names` and `cirq.ParallelGate`

vtomole:
I think we should do `cirq>=$thelatestversion` from now on. In this case `cirq>=0.13.0`. 
Thoughts?

richrines1:
> I think we should do cirq>=$thelatestversion from now on. In this case cirq>=0.13.0.
Thoughts?

in general I think it's best to keep the requirements loose whenever possible and only update lower bounds when we have a reason to, so as to avoid unnecessarily creating dependency issues for someone updating cirq-superstaq. That said, I think there are pretty good reasons to require `>=0.13` at this point

richrines1:
lgtm, leaving for others

dowusu-antwi (approved):
LGTM.

I'm sympathetic to Rich's comments -- it makes sense to be as loose as possible with requirements. But yeah, in this particular case might as well require `>=0.13`.

PR #146, opened 2021-12-28
Check cleanup

### LINKED ISSUE #145: check/pylint only checks files in examples/
`check/pylint` runs the line:
```
pylint --rcfile=.pylintrc examples cirq-superstaq
```
However `cirq-superstaq` (hyphenated) is not a directory, so pylint has only been seeing files in `examples`. Changing it to `cirq_superstaq` (underscore) brings up a couple line-too-long errors that had been evading our checks
```
************* Module cirq_superstaq.service
cirq_superstaq/service.py:272:0: C0301: Line too long (102/100) (line-too-long)
cirq_superstaq/service.py:297:0: C0301: Line too long (106/100) (line-too-long)
```
I'm not sure why these haven't been flagged by either `check/format` or `check/flake8`

richrines1:
* use `setup.cfg` for all check configuration
* forward command line arguments to checks
* use `install_types` and `non_interactive` flags to install missing stubs from typeshed via mypy
* fixes #145


check/all
```diff
@@ -11,12 +11,12 @@
 cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd "$(git rev-parse --show-toplevel)"
 
-# tests are ordered based on importance (e.g. when flake8 fails, it often means
+# tests are ordered based on importance (e.g. when format fails, it often means
 # there is a syntax error) and speed (slow checks last)
+echo "Running format check"
+check/format || exit 1  # this syntax causes whole script to terminate if format fails
```
dowusu-antwi, line 17: Why do we only include this termination for `format` and not, for example, for `flake8`?

singular-value: `./check/format` is somewhat unique because it also has the corresponding `./check/format --apply` script.

Running that script can often incidentally fix flake8 and pylint errors, so its nice to have a workflow of "wait does format work? If not, let's autofix before proceeding"

singular-value (commented):
Nice! Leaving for @dowusu-antwi and @vtomole 
And I think you said you can do this for qiskit-superstaq (and cirq-superstaq) too?

check/all
dowusu-antwi: Ah, I see (also explains why we've swapped the `format` check with the `flake8` check to run before it).

richrines1: it definitely makes sense for `format`, but I wouldn't be opposed to terminating on other failures as well

richrines1:
(just updated to make `service_test.py` not dependent on having environment variables unset)

> And I think you said you can do this for qiskit-superstaq (and cirq-superstaq) too?

yep will push those as soon as this one is reviewed & merged

check/all
richrines1: thoughts @vtomole @perlinm ?

vtomole: I agree.

perlinm: Hmm tbh I'm usually annoyed by a `format` failure blocking other tests, but I don't really mind either way.

perlinm: Actually: what if, rather than terminating `check/all`, we had a clear (and possibly colorful) FAILURE / SUCCESS message printed at the end?

vtomole: The checks exist as they are because of [KISS](https://en.wikipedia.org/wiki/KISS_principle). If it seems like we are slowly reinventing [tox](https://tox.wiki/en/latest/), we might as well use tox.

perlinm: `tox` seems to do much more than what we are asking here, but I wouldn't be opposed to using it if it looks like the right thing to use.  That said, I can't say I'm super needing it right now, so terminating upon failure is fine I guess.

richrines1: a simple compromise (see update): `check/all` terminates on failure, but `check/all -f` will keep going (there have been a few times that I've known a certain error was going to happen but still wanted to run everything)

vtomole (approved):
LGTM

PR #147, opened 2021-12-30
rxy qasm -> "rphi"

richrines1:


singular-value:
adding @perlinm as a reviewer

perlinm:
What's the reason for this change?  Also, why is it specified in QASM by half turns?  That sounds more like a `PhasedXPowGate`, whereas `Rxy` should resemble the `Rx` gate ([example](https://www.quantum-inspire.com/kbase/rx-gate/)).

richrines1:
> What's the reason for this change? 

in qiskit's version of QASM, `"rxx"` and `"rzz"` are used for two-qubit XX and ZZ interactions, and accordingly we use `"rzx"` for `css.ZXPowGate`. Following this pattern `"rxy"` would be a two-qubit XY interaction, not a single-qubit rotation around a variable axis. `"rphi"` was taken from e.g. eq. (1) in https://arxiv.org/pdf/1603.03996.pdf, but definitely open to suggestions if another string makes more sense

> Also, why is it specified in QASM by half turns? That sounds more like a `PhasedXPowGate`, whereas `Rxy` should resemble the `Rx` gate, which is specified by radians

`half_turns` is a cirq thing that tells it to format the string as `"rxy(pi*<half_turns>)"`; it's still specified in radians. (`cirq.Rx` uses it as well)

perlinm:
Ah I see.  In that case `rphi` sounds good to me.

perlinm (approved):
lgtm

PR #148, opened 2022-01-03
Add client-side functionality for setting IBMQ token

dowusu-antwi:
Lets `css` work with `PostIBMQToken` endpoint.


dowusu-antwi:
Appear to also be some errors with the way we're looking at the JSON storing the input token... looking into now.
client-set-error

richrines1:
shouldn't `ibmq_set_token()` already be inherited from `UserConfig`?

dowusu-antwi:
> shouldn't `ibmq_set_token()` already be inherited from `UserConfig`?

Yes -- I see now how it matches our use of `aqt_upload_configs`, also inherited from 'UserConfig'. Fixed.

dowusu-antwi:
Requesting re-review. `JSONDecodeError` fixed on SuperstaQ:

flask-fixes

dowusu-antwi:
Requesting re-review @richrines1 @vtomole.

vtomole (approved):
LGTM

PR #149, opened 2022-01-05
add __pow__ method to ParallelRxy gate class

perlinm:
... so that if `type(gate) is ParallelRxy`, then `type(gate ** power) is ParallelRxy`, which was not previously the case (and led to a bug in the CQDevice demo).


singular-value (commented):
lgtm, but leaving for others

richrines1 (approved):
lgtm

PR #150, opened 2022-01-06
Rename Rxy to Rphi

perlinm:


vtomole (approved):
LGTM. Thanks for the small doc cleanups as well!

richrines1:
what do you think about using `phi` instead of `axis_angle` now? I kind of prefer the more descriptive name but it feels especially weird to use "phi" in the name of we call the variable something else

perlinm:
> what do you think about using `phi` instead of `axis_angle` now? I kind of prefer the more descriptive name but it feels especially weird to use "phi" in the name of we call the variable something else

Hmm, this makes sense I suppose.  Should I also rename `rot_angle` to `theta`?

richrines1:
> Should I also rename `rot_angle` to `theta`?

yeah I guess that would be consistent (and also consistent w/ `FermionicSWAPGate`)

perlinm:
Alright done @richrines1

richrines1 (approved):
cool lgtm

PR #162, opened 2022-01-24
"Emit `num_copies` in` ParallelRGate`s QASM

vtomole:


vtomole:
Omg. the `num_copies` is first in QASM. hold on.

richrines1:
why this change? the qasm string for qiskit's `GR` gate doesn't include a `num_copies` parameter so to be consistent this shouldn't either imo

vtomole:
Oh, right.

vtomole:
Closing.

PR #168, opened 2022-01-26
Fixes integration test + Adds pulse parameter

dowusu-antwi:
Recent deployment of `applications-superstaq` requires updating the `requirements.txt` file to fix the integration test:

fix-integration

In particular we changed the `SuperstaQException` object, which I think is the source of the error. I was also working on passing in a parameter to explicitly request a pulse-enabled backend, so I just used this branch for the update. For example:
try-pulse-sim


singular-value (approved):
discussed in person and looks good

PR #171, opened 2022-02-02
FermionicSWAPGate -> ZZSwapGate

richrines1:
gate renamed to avoid conflicts w/ prior usage of "Fermionic SWAP"

parity: SupertechLabs/qiskit-superstaq#100


singular-value (approved):
lgtm

PR #174, opened 2022-02-07
Edited `cq_compile` circuit

### LINKED ISSUE #173: Integration test failed

dowusu-antwi:
Fixes #173. ~Added qubit measurement on `cq_compile` test. Whoops, misread the circuit -- the problem wasn't a measurement, but the last `cirq.Z` gate in the hard-coded circuit was applied to the wrong qubit.~ The compiled circuit was a bit different. I'm not sure if this is because of recent changes to `cq_compile` or not.

daily-integration

### DIFF
--- a/cirq_superstaq/daily_integration_test.py
+++ b/cirq_superstaq/daily_integration_test.py
@@ -100,8 +100,7 @@ def test_cq_compile(service: cirq_superstaq.Service) -> None:
     circuit = cirq.Circuit(
         cirq.H(qubits[0]),
         cirq.CNOT(qubits[0], qubits[1]),
-        cirq.measure(qubits[0]),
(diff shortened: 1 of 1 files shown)


vtomole (approved):
> because of recent changes to cq_compile or not.

It it was.

Thank you.

perlinm:
I think the last Z gate preceding the measurement should be eliminated entirely by the CQDevice compiler.  I should look into that and fix it (which will in turn require changing this integration test again).

dowusu-antwi:
> I think the last Z gate preceding the measurement should be eliminated entirely by the CQDevice compiler. I should look into that and fix it (which will in turn require changing this integration test again).

Ok sounds good, I'll just merge for now in favor of not having breaking code and can address the next integration test issue with another PR.

PR #177, opened 2022-02-09
sandwiching for AceCR

singular-value:
to support our Argonne users
working on parallel qss issue next


cirq_superstaq/custom_gates.py
```diff
@@ -219,43 +226,45 @@ def _decompose_(self, qubits: Tuple[cirq.LineQubit, cirq.LineQubit]) -> cirq.OP_
...
-        )
+        top, bottom = f"AceCR{self.polarity}(Z side)", f"AceCR{self.polarity}(X side)"
+        if self.sandwiched_gate is not None:
+            bottom += f"|{self.sandwiched_gate}|"
+        return cirq.protocols.CircuitDiagramInfo(wire_symbols=(top, bottom))
 
     def _qasm_(self, args: cirq.QasmArgs, qubits: Tuple[cirq.Qid, cirq.Qid]) -> Optional[str]:
-        """QASM symbol for AceCR('+-') (AceCR('-+')) is acecr_pm (acecr_mp)"""
+        """QASM symbol for AceCR('+-') (AceCR('-+')) is acecr_pm (acecr_mp).
+        Update me
```
singular-value, line 245: will do

cirq_superstaq/custom_gates.py
```diff
@@ -219,43 +226,45 @@ def _decompose_(self, qubits: Tuple[cirq.LineQubit, cirq.LineQubit]) -> cirq.OP_
...
         polarity_str = self.polarity.replace("+", "p").replace("-", "m")
-        return args.format(
-            "acecr_{0} {1},{2};\n",
-            polarity_str,
-            qubits[0],
-            qubits[1],
-        )
+        gate_name_qasm = f"acecr_{polarity_str}"
+        if self.sandwiched_gate is not None:
+            gate_name_qasm += f"_{self.sandwiched_gate}"
```
singular-value, line 250: @richrines1 / @vtomole what's the right way to handle the qasm?

Either we do what I've done here, where we'd have something like acecr_mp for non-sandwiched acecr and acecr_mp_Rx(0.5π) for a sandwiched one.

Or we always have the format acecr_polarity_sandwich, where the last piece could be "None".

cirq_superstaq/custom_gates_test.py
```diff
@@ -181,23 +208,39 @@ def test_acecr() -> None:
...
+    assert str(cirq_superstaq.AceCR("+-", cirq.X)) == "AceCR+-|X|"
 
 
 def test_acecr_decompose() -> None:
     a = cirq.LineQubit(0)
     b = cirq.LineQubit(1)
-    assert cirq.decompose_once(cirq_superstaq.AceCRMinusPlus(a, b)) is not None
-    cirq.testing.assert_decompose_is_consistent_with_unitary(
-        cirq_superstaq.AceCRMinusPlus, ignoring_global_phase=True
-    )
```
singular-value, line 200: Deleted this assertion. Don't think it is helpful since the unitary for AceCR is just inferred from the decomposition anyways.

cirq_superstaq/custom_gates.py
```diff
@@ -206,11 +206,18 @@ def __repr__(self) -> str:
 CR = ZX = ZXPowGate()  # standard CR is a full turn of ZX, i.e. exponent = 1
 
 
+@cirq.value_equality
 class AceCR(cirq.Gate):
-    def __init__(self, polarity: str) -> None:
-        assert polarity in ["+-", "-+"]
+    def __init__(self, polarity: str, sandwiched_gate: Optional[cirq.Gate] = None) -> None:
```
richrines1, line 211: imo given that the sandwiched gate can only be an X rotation, this argument should just be either an angle or an exponent (defaulting to 0)

cirq_superstaq/custom_gates.py
richrines1: > Either we do what I've done here, where we'd have something like acecr_mp for non-sandwiched acecr and acecr_mp_Rx(0.5π) for a sandwiched one.

imo this makes the most sense

cirq_superstaq/custom_gates.py
singular-value: ah good 👍 will update

cirq_superstaq/custom_gates.py
singular-value: addressed

cirq_superstaq/custom_gates.py
singular-value: done

cirq_superstaq/custom_gates.py
```diff
@@ -206,11 +206,31 @@ def __repr__(self) -> str:
...
+
+    The typical AceCR in literature is a positive half-CR, then X on control, then negative half-CR.
+    Args:
+        polarity: should be either "+-" or "-+". Specifies if positive or negative half-CR is first
+        sandwich_rx_rads: during the X on control, an rx gate can be simultaneously
+            applied to the target.
+    """
+
+    def __init__(
+        self, polarity: str, sandwich_rx_rads: Optional[cirq.value.TParamVal] = None
```
richrines1, line 221: why not default to zero instead of `None`? (and elsewhere replace `if sandwich_rx_rads is None: ...` with `if sandwich_rx_rads: ...`)

cirq_superstaq/custom_gates.py
```diff
@@ -219,43 +239,49 @@ def _decompose_(self, qubits: Tuple[cirq.LineQubit, cirq.LineQubit]) -> cirq.OP_
...
+        If there is a sandwich, it comes last. For example, AceCR("-+", np.pi / 2) has qasm
+        acecr_mp_rx(pi*0.5).
+        """
+        polarity_str = self.polarity.replace("+", "p").replace("-", "m")
+        gate_name_qasm = f"acecr_{polarity_str}"
+        kwargs: dict = {"q0": qubits[0], "q1": qubits[1]}
+        if self.sandwich_rx_rads is not None:
+            kwargs["theta"] = self.sandwich_rx_rads / np.pi
+            gate_name_qasm += "_rx({theta:half_turns})"
+        return args.format(gate_name_qasm + " {q0},{q1};\n", **kwargs)
```
richrines1, line 268: can be simplified a bit:
```suggestion
        if not self.sandwich_rx_rads:
            return args.format("acecr_{} {},{};\n", polarity_str, *qubits)
        exponent = self.sandwich_rx_rads / np.pi
        return args.format("acecr_{}_rx({:half_turns}) {},{};\n", polarity_str, exponent, *qubits)
```

cirq_superstaq/custom_gates.py
```diff
@@ -206,11 +206,31 @@ def __repr__(self) -> str:
...
+    """
+
+    def __init__(
+        self, polarity: str, sandwich_rx_rads: Optional[cirq.value.TParamVal] = None
+    ) -> None:
+        assert polarity in ["+-", "-+"], "Polarity must be +- or -+"
         self.polarity = polarity
 
+        if sandwich_rx_rads is not None:
+            assert sandwich_rx_rads in [
```
richrines1, line 227: should we normalize `sandwich_rx_rads` so that this accepts e.g. -π or 3π/2?

cirq_superstaq/custom_gates.py
```diff
@@ -219,43 +239,49 @@ def _decompose_(self, qubits: Tuple[cirq.LineQubit, cirq.LineQubit]) -> cirq.OP_
...
+        kwargs: dict = {"q0": qubits[0], "q1": qubits[1]}
+        if self.sandwich_rx_rads is not None:
+            kwargs["theta"] = self.sandwich_rx_rads / np.pi
+            gate_name_qasm += "_rx({theta:half_turns})"
+        return args.format(gate_name_qasm + " {q0},{q1};\n", **kwargs)
 
-    def __hash__(self) -> int:
-        return hash(self.polarity)
+    def _value_equality_values_(self) -> Any:
+        return (self.polarity, self.sandwich_rx_rads)
```
richrines1, line 271: nice

cirq_superstaq/custom_gates.py
richrines1: actually what do you think about not restricting the angle at all? the gate works valid for any x-rotation, it's just particular hardware that advantages the set `{0, ±π/2, π}` - so restricting to this set should maybe be relegated to the compiler and not the gate itself

singular-value: 👍 good point --> we will eventually support single-pulse rx(theta) anyways

cirq_superstaq/custom_gates.py
singular-value: 👍

cirq_superstaq/custom_gates.py
```diff
@@ -206,10 +206,22 @@ def __repr__(self) -> str:
...
 class AceCR(cirq.Gate):
-    def __init__(self, polarity: str) -> None:
-        assert polarity in ["+-", "-+"]
+    """Active Cancellation Echoed Cross Resonance gate, supporting polarity switches and sandwiches.
+
+    The typical AceCR in literature is a positive half-CR, then X on control, then negative half-CR.
+    Args:
+        polarity: should be either "+-" or "-+". Specifies if positive or negative half-CR is first
+        sandwich_rx_rads: during the X on control, an rx gate can be simultaneously
+            applied to the target.
```
richrines1, line 217: nit: should indicate that this argument is the angle of that rx() gate

cirq_superstaq/custom_gates.py
```diff
@@ -206,10 +206,22 @@ def __repr__(self) -> str:
...
+    """
+
+    def __init__(self, polarity: str, sandwich_rx_rads: Optional[cirq.value.TParamVal] = 0) -> None:
```
richrines1, line 220:
```suggestion
    def __init__(self, polarity: str, sandwich_rx_rads: cirq.TParamVal = 0) -> None:
```
(`Optional[A]` is just an alias for `Union[A, None]`, so not necessary here)

cirq_superstaq/custom_gates.py
```diff
@@ -219,43 +231,47 @@ def _decompose_(self, qubits: Tuple[cirq.LineQubit, cirq.LineQubit]) -> cirq.OP_
...
+        """
+        polarity_str = self.polarity.replace("+", "p").replace("-", "m")
+        if not self.sandwich_rx_rads:
+            return args.format("acecr_{} {},{};\n", polarity_str, *qubits)
+        exponent = self.sandwich_rx_rads / np.pi
+        return args.format("acecr_{}_rx({:half_turns}) {},{};\n", polarity_str, exponent, *qubits)
 
-    def __hash__(self) -> int:
-        return hash(self.polarity)
+    def _value_equality_values_(self) -> Any:
```
richrines1, line 260:
```suggestion
    def _value_equality_values_(self) -> Tuple[str, cirq.TParamVal]:
```

richrines1 (commented):
given that the `sandwich_rx_rads` argument is annotated as possibly being parameterized, it might be good to add the method:
```
    def _parameter_names_(self) -> AbstractSet[str]:
        return cirq.parameter_names(self.sandwich_rx_rads)
```
(which should make `cirq.parameter_names(gate)` and `cirq.is_parameterized(gate)` work correctly)

cirq_superstaq/custom_gates.py
singular-value: 👍

singular-value:
@richrines1 I ended up backtracking and got rid of support for symbolic parameters for the rx angle. Figured we could keep it as future work if it is needed.

cirq_superstaq/custom_gates.py
```diff
@@ -202,10 +202,22 @@ def __repr__(self) -> str:
 CR = ZX = ZXPowGate()  # standard CR is a full turn of ZX, i.e. exponent = 1
 
 
+@cirq.value_equality
 class AceCR(cirq.Gate):
-    def __init__(self, polarity: str) -> None:
-        assert polarity in ["+-", "-+"]
+    """Active Cancellation Echoed Cross Resonance gate, supporting polarity switches and sandwiches.
+
+    The typical AceCR in literature is a positive half-CR, then X on control, then negative half-CR.
```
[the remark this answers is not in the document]
singular-value: Nope -->updating

cirq_superstaq/custom_gates.py
```diff
@@ -215,43 +227,47 @@ def _decompose_(self, qubits: Tuple[cirq.LineQubit, cirq.LineQubit]) -> cirq.OP_
...
 
     def _circuit_diagram_info_(
         self, args: cirq.CircuitDiagramInfoArgs
     ) -> cirq.protocols.CircuitDiagramInfo:
-        return cirq.protocols.CircuitDiagramInfo(
-            wire_symbols=(f"AceCR{self.polarity}(Z side)", f"AceCR{self.polarity}(X side)")
-        )
+        top, bottom = f"AceCR{self.polarity}(Z side)", f"AceCR{self.polarity}(X side)"
+        if self.sandwich_rx_rads:
+            bottom += f"|{cirq.rx(self.sandwich_rx_rads)}|"
```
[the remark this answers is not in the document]
singular-value: I see. Reflecting on it a bit, I prefer the first option.

singular-value:
(updated https://github.com/SupertechLabs/qiskit-superstaq/pull/105 to keep docstrings in sync)

cirq_superstaq/custom_gates.py
```diff
@@ -215,43 +228,47 @@ def _decompose_(self, qubits: Tuple[cirq.LineQubit, cirq.LineQubit]) -> cirq.OP_
...
+        polarity_str = self.polarity.replace("+", "p").replace("-", "m")
+        if not self.sandwich_rx_rads:
+            return args.format("acecr_{} {},{};\n", polarity_str, *qubits)
+        exponent = self.sandwich_rx_rads / np.pi
+        return args.format("acecr_{}_rx({:half_turns}) {},{};\n", polarity_str, exponent, *qubits)
 
-    def __hash__(self) -> int:
-        return hash(self.polarity)
+    def _value_equality_values_(self) -> Tuple[str, float]:
+        return (self.polarity, self.sandwich_rx_rads)
```
richrines1, line 258: just realizing that this doesn't handle the periodicity of `sandwich_rx_rads`, so `==`, `cirq.approx_eq`, and `cirq.equal_up_to_global_phase` won't always work as expected

`approx_eq` and `==` could be fixed with,
```suggestion
    def _value_equality_values_(self) -> Tuple[str, cirq.PeriodicValue]:
        return self.polarity, cirq.PeriodicValue(self.sandwich_rx_rads, 4 * np.pi)
```
where the period is 4π instead of 2π because rx(θ) and rx(θ+2π) differ by a global phase. Not sure what difference it would make in this case, but fwiw gates defined in cirq instead implement something like,
```suggestion
    def _value_equality_values_(self) -> Tuple[str, float]:
        return self.polarity, self.sandwich_rx_rads % (4 * np.pi)

    def _value_equality_approximate_values_(self) -> Tuple[str, cirq.PeriodicValue]:
        return self.polarity, cirq.PeriodicValue(self.sandwich_rx_rads, 4 * np.pi)
```
which (I think) should work the same way in every case

To get `equal_up_to_global_phase(AceCR("+-", θ), AceCR("+-", θ+2π))` to work we'd also need to add something like,
```
    def _equal_up_to_global_phase_(self, other, atol): 
        if isinstance(other, AceCR):
            if other.polarity == self.polarity and cirq.all_near_zero_mod(
                other.sandwich_rx_rads - self.sandwich_rx_rads, 2 * np.pi, atol=atol
            ):
                return True
            return False
        return NotImplemented
```
however this function isn't necessary if the `cirq.rx` in `_decompose_` can be replaced with with `cirq.X ** (self.sandwich_rx_rads / (2 * np.pi))`, in which case the periods in the above suggestions should just be 2π instead of 4π

richrines1: or maybe way simpler: is you add something like
```
        sandwich_rx_rads = cirq.ops.fsim_gate._canonicalize(sandwich_rx_rads)
```
to `__init__`, then I believe everything will work as expected if you leave `_value_equality_values_` as it is and add the one function,
```
    def _value_equality_approximate_values_(self) -> Tuple[str, cirq.PeriodicValue]:
        return self.polarity, cirq.PeriodicValue(self.sandwich_rx_rads, 2 * np.pi)
```

singular-value: 👍 did the second suggestion

cirq_superstaq/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.1.12"
+__version__ = "0.1.13"
```
richrines1, line 1: do you know why this and the `ibmq_pulse` stuff is showing up here? this branch seems to be up to date with `main`..

singular-value: oops bad git usage

cirq_superstaq/custom_gates.py
```diff
@@ -202,10 +202,26 @@ def __repr__(self) -> str:
...
+            with the X gate on the "Z side".
+    """
+
+    def __init__(self, polarity: str, sandwich_rx_rads: float = 0) -> None:
+        if polarity not in ("+-", "-+"):
+            raise ValueError("Polarity must be either '+-' or '-+'")
         self.polarity = polarity
+        self.sandwich_rx_rads = cirq.ops.fsim_gate._canonicalize(sandwich_rx_rads)
+
+    def _value_equality_approximate_values_(self) -> Tuple[str, cirq.PeriodicValue]:
```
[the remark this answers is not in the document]
singular-value: good call thanks. Updated

cirq_superstaq/custom_gates.py
```diff
@@ -202,10 +202,23 @@ def __repr__(self) -> str:
...
 
 
+@cirq.value_equality(approximate=True)
 class AceCR(cirq.Gate):
-    def __init__(self, polarity: str) -> None:
-        assert polarity in ["+-", "-+"]
+    """Active Cancellation Echoed Cross Resonance gate, supporting polarity switches and sandwiches.
+
+    The typical AceCR in literature is a positive half-CR, then X on "Z side", then negative
+    half-CR.  ("Z side" and "X side" refer to the two sides of the underlying ZX interactions).
```
richrines1, line 210:
```suggestion
    half-CR ("Z side" and "X side" refer to the two sides of the underlying ZX interactions).
```

richrines1 (dismissed):
lg w/ one more nit

cirq_superstaq/custom_gates.py
richrines1: oh wait hold on

singular-value:
@richrines1 re-requesting review

richrines1 (approved):
lgtm!

singular-value:
Thanks for the thorough review!

singular-value:
Not sure what happened here --> "Files changed" shows that only 3 files were changed (a notebook and custom_gates{_test}.py).

But the `main` branch is showing that _version.py and service.py were changed too. 🤔

richrines1:
weird yeah it looks like #178 and #179 were reverted

cirq_superstaq/_version.py
dowusu-antwi: ^ what does this refer to? From live discussion with @vtomole, might be where the problem comes from.

PR #182, opened 2022-02-17
Bring back PRs 178 and 179

singular-value:
Somehow got undone by #177


dowusu-antwi (approved):

vtomole:
How so? The PR you linked doesn't have a diff for `_version.py`.

dowusu-antwi:
Somehow there was a commit (https://github.com/SupertechLabs/cirq-superstaq/commit/1c28c0905b1e3623ee13a53a96ef33f4c11b5409) that reverted `_version.py`, even though it isn't in the latest diff.

The reversion still shows up in the `acecr-upgrade` branch strangely enough.

singular-value:
Yes, this is the mystery. I'll buy a drink for anyone who figures it out.

vtomole:
> The reversion still shows up in the acecr-upgrade branch strangely enough.

@dowusu-antwi So basically the problem is that the branch wasn't properly reflected in Github'a diff. What happens if you locally diff `acecr-upgrade` vs main?

PR #185, opened 2022-02-21
pip install third-party reqs in examples

richrines1:


vtomole:
Some people might have qiskit on their system already. Let's do

```python
try:
    import qiskit
except ImportError:
    print("installing qiskit...")
    !pip install -q qiskit-terra==0.18.3
    print("installed qiskit.")
```

richrines1:
@vtomole updated (also had to include a check for the correct versions)

vtomole:
I forget. The `;` after the `out.draw()` used to be necessary, if i recall the plot would be drawn twice otherwise. Looks like this was fixed?

vtomole (approved):

PR #193, opened 2022-03-03
retire casablanca

singular-value:


singular-value:
Fixes #192 
(ran daily integration from this branch and it passed: https://github.com/SupertechLabs/cirq-superstaq/actions/runs/1929475193

vtomole (approved):

PR #209, opened 2022-03-24
Update to v0.1.17

stephanielee9:


vtomole (approved):

vtomole (changes requested):
@stephanielee9 I'm looking into why this check is failing. Let's not merge until then.

vtomole (approved):
Cirq version incompatibility fixed by https://github.com/SupertechLabs/cirq-superstaq/pull/211

Issue #51, opened 2021-09-03
Catch `ModuleNotFoundError`s and throw  more specific exception


vtomole:
@singular-value yeah I'd just set it up to catch `ModuleNotFoundError`s and throw something else - will add it now

_Originally posted by @richrines1 in https://github.com/SupertechLabs/cirq-superstaq/issues/50#issuecomment-912671484_

richrines1:
this ended up being addressed in #50, however currently the error only gets thrown after receiving an unusable result back from the server. eventually it might be better to preempt the server call?

singular-value:
@richrines1 I'm not sure what you mean, can you elaborate?

richrines1:
@singular-value basically the current sequence is something like: 
1. user calls `ibmq_compile()`
2. the client posts a job to the server, which returns a compiled sequence
3. the client throws a `ModuleNotFoundError` when it isn't able unpack the sequence it got back
4. we catch the error and throw a more informative `SupererstaQModuleNotFoundException`

it seems like it would be better to bypass step 2 (e.g. by checking for modules explicitly instead of waiting for `ModuleNotFoundError`s), so the client doesn't have to post a job and get a result back just to find out it can't use it. so basically:
1. user calls `ibmq_compile()`
2. the client checks for the necessary modules (qiskit in this case)
3. the client throws a `SupererstaQModuleNotFoundException`

vtomole:
I agree.

singular-value:
👍 makes sense

richrines1:
i think this was covered/superseded by #219

PR #219, opened 2022-04-08
check for qiskit in ibmq_compile

richrines1:
related to #51, but now that `ibmq_compile` returns both pulse schedules and compiled circuits we can throw a warning but still return the circuits when qiskit isn't installed

also fixes: #200

### DIFF
--- a/cirq_superstaq/compiler_output_test.py
+++ b/cirq_superstaq/compiler_output_test.py
@@ -44,16 +44,18 @@ def test_read_json_ibmq() -> None:
     assert not hasattr(out, "circuit")
     assert not hasattr(out, "pulse_sequence")
 
-    with mock.patch("qiskit.__version__", "0.17.2"), pytest.raises(
-        applications_superstaq.SuperstaQException, match="Qiskit Terra version"
+    with mock.patch.dict("sys.modules", {"qiskit": None}), pytest.warns(
+        UserWarning, match="requires Qiskit Terra"
     ):
-        _ = cirq_superstaq.compiler_output.read_json_ibmq(json_dict, circuits_is_list=False)
-
-    with mock.patch.dict("sys.modules", {"qiskit": None}):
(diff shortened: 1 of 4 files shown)


cirq_superstaq/service.py
```diff
@@ -334,33 +334,20 @@ def ibmq_compile(
...
-        return cirq_superstaq.compiler_output.CompilerOutput(
-            circuits=compiled_circuits, pulse_sequences=pulses
-        )
+
+        return cirq_superstaq.compiler_output.read_json_ibmq(json_dict, circuits_is_list)
 
     def neutral_atom_compile(
         self, circuits: Union[cirq.Circuit, List[cirq.Circuit]], target: str = "neutral_atom_qpu"
-    ) -> Any:
+    ) -> cirq_superstaq.compiler_output.CompilerOutput:
```
vtomole, line 350: We don't return ` cirq_superstaq.compiler_output.CompilerOutput` in this function but maybe this is something we want to create an issue for. 

Anyway, i'm surprised that this is passing since we return ` pulses = applications_superstaq.converters.deserialize(json_dict["pulses"])`

vtomole:
LGTM once modulo my question.

cirq_superstaq/service.py
richrines1: good catch. fixed and created #221

vtomole (approved):

cirq_superstaq/compiler_output_test.py
```diff
@@ -58,18 +95,18 @@ def test_read_json() -> None:
...
 
     jaqal_program = textwrap.dedent(
         """\
-                register allqubits[1]
-                prepare_all
-                R allqubits[0] -1.5707963267948966 1.5707963267948966
-                Rz allqubits[0] -3.141592653589793
-                measure_all
-                """
+        register allqubits[1]
```
vtomole, line 104: Link

PR #222, opened 2022-04-08
support singular `CompilerOutput.jaqal_program`

richrines1:
aligns plurality of `CompilerOutput.jaqal_program(s)` with that of `.circuit(s)` (in line with `qiskit_superstaq.CompilerOutput`)

also corrects `CompulerOutput.__repr__`, however this could definitely use an overhaul (see #103)

### DIFF
--- a/cirq_superstaq/daily_integration_test.py
+++ b/cirq_superstaq/daily_integration_test.py
@@ -98,17 +98,17 @@ def test_qscout_compile(service: cirq_superstaq.Service) -> None:
 
     jaqal_program = textwrap.dedent(
         """\
-                register allqubits[1]
+        register allqubits[1]
 
-                prepare_all
-                R allqubits[0] -1.5707963267948966 1.5707963267948966
-                Rz allqubits[0] -3.141592653589793
-                measure_all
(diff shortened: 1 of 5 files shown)


cirq_superstaq/daily_integration_test.py
```diff
@@ -98,17 +98,17 @@ def test_qscout_compile(service: cirq_superstaq.Service) -> None:
 
     jaqal_program = textwrap.dedent(
         """\
-                register allqubits[1]
+        register allqubits[1]
```
vtomole, line 101: Might need to merge `main` to this branch since this change is there: https://github.com/SupertechLabs/cirq-superstaq/pull/219/files#r846320292

vtomole (approved):

cirq_superstaq/daily_integration_test.py
richrines1: haha just spent a bit super confused because merging wasn't fixing this, but i think that one was just an identical change in `compiler_output.py`

vtomole (approved):

PR #223, opened 2022-04-08
update applications-superstaq req

richrines1:
necessary for the `aqt_save_configs` endpoint (corresponding qss: supertechlabs/qiskit-superstaq#143)

(we need to add a test for this)


vtomole:
> we need to add a test for this

Should we do that in this PR or merge this and file and issue?

richrines1:
let's merge this first. created supertechlabs/applications-superstaq#48

vtomole (approved):

Issue #210, opened 2022-03-25
`test_parallel_gates_equivalence_groups()` fails with new Cirq update


vtomole:
https://github.com/SupertechLabs/cirq-superstaq/runs/5683094136?check_suite_focus=true

vtomole:
Apprently

```python
import cirq
import cirq_superstaq

res = cirq_superstaq.ParallelGates(cirq.X, cirq_superstaq.ZZSwapGate(1.23), cirq.X).on(cirq.LineQubit(0), cirq.LineQubit(1), cirq.LineQubit(2), cirq.LineQubit(3))
res1 = cirq_superstaq.ParallelGates(cirq.X, cirq_superstaq.ZZSwapGate(1.23), cirq.X).on(cirq.LineQubit(3), cirq.LineQubit(1), cirq.LineQubit(2), cirq.LineQubit(0))

print(res)
print(res1)
print(res == res1)
```

Should be `True`, but now it's `False`

vtomole:
Taking this as @richrines1 is busy.

vtomole:
We've merged https://github.com/SupertechLabs/cirq-superstaq/pull/211 temporarily until this issue is closed.

richrines1:
this issue was fixed in cirq 0.15. We added a workaround so that tests still pass using cirq 0.14.x (#237), but technically the issue still exists for users using that version

that said imo we can close this because (1) it only comes up in really niche cases, (2) the issue is on cirq's end so there doesn't seem to be an easy workaround, and (3) even if there was we'll probably update our requirements to `cirq>=0.15.0` before we get around to fixing it. @vtomole?

vtomole:
I agree that this is good to close.

PR #227, opened 2022-04-13
increase cirq requirement upper bound

richrines1:
(because #210 doesn't affect some pre-release versions of `cirq` 0.14.0)


vtomole (approved):

vtomole (dismissed):

vtomole:
We can go even higher on bounding the dev version. https://pypi.org/project/cirq/0.14.0.dev20220203001129/ is the newest that doesn't break our tests.

vtomole:
> Cannot instantiate abstract class "Result" with abstract attributes "data", "measurements" and "params"  [abstract]

The new thing is [ResultDict](https://quantumai.google/reference/python/cirq/study/ResultDict).

cirq_superstaq/custom_gates_test.py
```diff
@@ -216,7 +216,7 @@ def test_acecr_eq() -> None:
     assert cirq_superstaq.AceCR("-+", np.pi) != cirq_superstaq.AceCR("+-", np.pi)
 
     assert cirq_superstaq.AceCR("+-", np.pi) == cirq_superstaq.AceCR("+-", 5 * np.pi)
-    assert cirq_superstaq.AceCR("+-", np.pi) != cirq_superstaq.AceCR("+-", 3 * np.pi)
+    assert cirq_superstaq.AceCR("+-", np.pi) == cirq_superstaq.AceCR("+-", 3 * np.pi)
```
richrines1, line 219: this was previously only False due to a (now corrected) quirk in `cirq.ops.fsim_gate._canonicalize()` where `_canonicalize(pi+0.1) ==  _canonicalize(3*pi+0.1)` but `_canonicalize(pi) !=  _canonicalize(3*pi)`

do we want it to be False generally? we use `cirq.rx` for the sandwiched gate, so technically `AceCR("+-", π)` and `AceCR("+-", 3π)` should differ by a global phase. cirq would typically make these not equal (e.g. `cirq.rx(π) != cirq.rx(3π)`

alternatively we could replace `cirq.rx` in `_decompose_` with a `cirq.XPowGate` so that the global phase doesn't change between `AceCR(..., ϕ)` and `AceCr(..., ϕ+2π)`

vtomole, line 219: Why the change?

richrines1: this was previously only False due to a (now corrected) quirk in `cirq.ops.fsim_gate._canonicalize()` where `_canonicalize(pi+0.1) ==  _canonicalize(3*pi+0.1)` but `_canonicalize(pi) !=  _canonicalize(3*pi)`

do we want it to be False generally? we use `cirq.rx` for the sandwiched gate, so technically `AceCR("+-", π)` and `AceCR("+-", 3π)` should differ by a global phase. cirq would typically make these not equal (e.g. `cirq.rx(π) != cirq.rx(3π)`

alternatively we could replace `cirq.rx` in `_decompose_` with a `cirq.XPowGate` so that the global phase doesn't change between `AceCR(..., ϕ)` and `AceCr(..., ϕ+2π)`

richrines1:
> > Cannot instantiate abstract class "Result" with abstract attributes "data", "measurements" and "params"  [abstract]
> 
> Replacing `Result` with [ResultDict](https://quantumai.google/reference/python/cirq/study/ResultDict) will solve.

will this work for all `cirq>=0.13.0`?

vtomole:
> will this work for all cirq>=0.13.0?

Sadly no, this was released in `0.14.0`.

cirq_superstaq/custom_gates_test.py
vtomole: Is the question more generally whether our equality check should take into account global phase?

vtomole: Discussed offline: Delaying doing this for later to avoid potentially breaking ongoing experiments.

richrines1: > Is the question more generally whether our equality check should take into account global phase?

(basically yes. cirq's equality check does for most gates, but we currently don't for AceCR)

richrines1:
> Let's try https://pypi.org/project/cirq/0.14.0.dev20220126174724/

that one seems to work!

vtomole (approved):

PR #234, opened 2022-04-20
Increase cirq uper bound version to `0.15.0.dev20220420201205`

vtomole:


vtomole:
@richrines1 Ready for review.

cirq_superstaq/service.py
```diff
@@ -66,7 +66,7 @@ def counts_to_results(
         for key in range(counts_of_key):
             samples.append(keys_as_list)
 
-    result = cirq.Result(
+    result = cirq.ResultDict(
```
richrines1, line 69: will this will break for `cirq<0.14.0`?

vtomole: What do you think about abandoning `<0.14.0` now?

```python
 from packaging import version
    if version.parse(cirq.__version__) < version.parse("0.14.0"):
        result = cirq.Result(
            params=param_resolver,
            measurements={
                combine_key_names: np.array(samples),
            },
        )
    else:
        result = cirq.ResultDict(
            params=param_resolver,
            measurements={
                combine_key_names: np.array(samples),
            },
        )
```

Looks a bit.. i don't' know.. We'll have to install 2 versions of Cirq on Github Actions to touch all the lines.

richrines1: what about
```
ResultDict = (
    cirq.Result if version.parse(cirq.__version__) < version.parse("0.14.0") else cirq.ResultDict
)

result = ResultDict(
    ...
)
```

vtomole: Mypy flags that the "other Result" type doesn't exist.

richrines1: ah right. one other idea:
```
ResultDict = getattr(cirq, "ResultDict", getattr(cirq, "Result"))
```

vtomole: V cleva. Pushed and opened: https://github.com/SupertechLabs/cirq-superstaq/issues/236

richrines1:
do we still have to update the `ParallelGates` tests so that they pass w/ cirq 0.14.x?

vtomole:
> do we still have to update the ParallelGates tests so that they pass w/ cirq 0.14.x?

Is there a way we can do this short of monkey patching? We can encourage cirq-supertstaq devs to use the latest version of Cirq and have this be a bug in cirq-superstaq. Basically, i wonder how much ParrallelGates equality checks will be done.

richrines1 (approved):
lgtm. just confirmed that everything passes w/ both cirq==0.13.0 and cirq==0.14.1

PR #241, opened 2022-04-28
Patch removal of `device` attribute from `cirq.Circuit()`

vtomole:
This PR gets around https://github.com/quantumlib/Cirq/pull/5189; which is a breaking change.


richrines1 (approved):
lgtm

PR #244, opened 2022-05-06
Use `cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent` for circuit comparison in `daily_integration_test.py`

vtomole:
Fixes https://github.com/SupertechLabs/cirq-superstaq/issues/243

### DIFF
--- a/cirq_superstaq/daily_integration_test.py
+++ b/cirq_superstaq/daily_integration_test.py
@@ -46,31 +46,24 @@ def test_acer_non_neighbor_qubits_compile(service: cirq_superstaq.Service) -> No
 def test_aqt_compile(service: cirq_superstaq.Service) -> None:
     qubits = cirq.LineQubit.range(8)
     circuit = cirq.Circuit(cirq.H(qubits[4]))
-    expected = cirq.Circuit(
-        cirq.rz(np.pi / 2)(qubits[4]),
-        cirq.rx(np.pi / 2)(qubits[4]),
-        cirq.rz(np.pi / 2)(qubits[4]),
-    )
 
     cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent(
-        service.aqt_compile(circuit).circuit, expected, atol=1e-08
+        service.aqt_compile(circuit).circuit, circuit, atol=1e-08
     )
 
     compiled_circuits = service.aqt_compile([circuit]).circuits
     assert isinstance(compiled_circuits, list)
     for compiled_circuit in compiled_circuits:
         cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent(
-            compiled_circuit, expected, atol=1e-08
+            compiled_circuit, circuit, atol=1e-08
         )
     compiled_circuits = service.aqt_compile([circuit, circuit]).circuits
 
     assert isinstance(compiled_circuits, list)
     for compiled_circuit in compiled_circuits:
         cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent(
-            compiled_circuit, expected, atol=1e-08
+            compiled_circuit, circuit, atol=1e-08
         )
-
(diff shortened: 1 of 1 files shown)


cirq_superstaq/daily_integration_test.py
```diff
@@ -51,8 +51,25 @@ def test_aqt_compile(service: cirq_superstaq.Service) -> None:
         cirq.rx(np.pi / 2)(qubits[4]),
         cirq.rz(np.pi / 2)(qubits[4]),
     )
-    assert service.aqt_compile(circuit).circuit == expected
-    assert service.aqt_compile([circuit]).circuits == [expected]
+
+    cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent(
+        service.aqt_compile(circuit).circuit, expected, atol=1e-08
```
richrines1, line 56: do we need `expected` anymore or could this just be
```suggestion
        service.aqt_compile(circuit).circuit, circuit, atol=1e-08
```

vtomole: How so? `expected` is different from `circuit`

`circuit = cirq.Circuit(cirq.H(qubits[4]))` vs 
 ```
expected = cirq.Circuit(
        cirq.rz(np.pi / 2)(qubits[4]),
        cirq.rx(np.pi / 2)(qubits[4]),
        cirq.rz(np.pi / 2)(qubits[4]),
    )```

richrines1: i believe `assert_circuits_with_terminal_measurements_are_equivalent` only compares unitaries, which should generally be equivalent

maybe this isn't what we want in this case though, if e.g. we're trying to test that we get the "optimized" hadamard decomposition (i.e. `rz`, `rx`, `rz`) instead of the less canonical (`rx`, `rz`, `rx`)

vtomole: Ah, not familiar with the guts assert_circuits_with_terminal_measurements_are_equivalent`. Should we revert this given that we are indeed trying to assert a certain optimization?

richrines1: ah maybe in that case? in most cases it seems pretty nice because it should be robust against future updates/optimizations (e.g. i can imagine there being subtle changes in the compiled version of the the more complex `test_cq_compile` circuit below) but in this one case maybe the circuit is simple enough that we shouldn't expect anything to change

or maybe another option is to keep `assert_circuits_with_terminal_measurements_are_equivalent`, and instead of comparing to a specific circuit add something that just checks that there's only one `rx` gate in the compiled circuit

vtomole: I'm fine with leaving as-is. Internal checks should make sure that things are being optimized appropriately.

richrines1: i agree. in that case i think `expected` definitely isn't necessary here & below

vtomole: Done.

cirq_superstaq/daily_integration_test.py
```diff
@@ -129,7 +141,9 @@ def test_cq_compile(service: cirq_superstaq.Service) -> None:
     )
 
     out = service.cq_compile(circuit)
-    assert out.circuit == compiled_circuit
+    cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent(
+        out.circuit, compiled_circuit, atol=1e-08
```
richrines1, line 145: this one should work too
```suggestion
        out.circuit, circuit, atol=1e-08
```

vtomole: Done.

richrines1 (dismissed):
lgtm

vtomole:
@richrines1 Please re-approve.

richrines1 (approved):

PR #248, opened 2022-05-18
Changing every instance of `import cirq_superstaq`  to `import cirq_superstaq as css`

pgoiporia:


vtomole (changes requested):
`sampler.py` still contains `import cirq_superstaq`

stephanielee9:
Wrapper, notebook, readme should have `css`

vtomole (changes requested):
All the docstrings should have `css` as well.

vtomole (approved):

PR #255, opened 2022-05-26
Update `qiskit-terra` version

vtomole:
Fixes: https://github.com/SupertechLabs/cirq-superstaq/issues/254


singular-value (approved):
lgtm but tagging @richrines1 for awareness

PR #257, opened 2022-05-26
make `Optional` not optional

richrines1:
closes: SupertechLabs/applications-superstaq#63 (along with SupertechLabs/qiskit-superstaq#168 and SupertechLabs/applications-superstaq#65)


vtomole (approved):
LGTM

PR #265, opened 2022-06-14
Added cirq_to_qiskit conversion function to Service

teaguetomesh:
Currently the function returns the _serialized_ `qiskit.QuantumCircuits` -- this is to avoid needing to import qiskit and qiskit-superstaq within cirq-superstaq. The deserialization can be done within SupermarQ.


richrines1:
we should add a check in `service.convert_to_qiskit` for qiskit's existence and version like what we have in `ibmq_compile`

dowusu-antwi:
(No longer needed, was for ISCA tutorial.)

PR #267, opened 2022-06-15
update some type annotations

richrines1:


cirq_superstaq/compiler_output.py
```diff
@@ -17,7 +17,7 @@ class CompilerOutput:
     def __init__(
         self,
         circuits: Union[cirq.Circuit, List[cirq.Circuit]],
-        pulse_sequences: Any = None,
+        pulse_sequences: Optional[Any] = None,
```
vtomole, line 20: Very interesting that the recent `Optional` flag that we added to our mypy config doesn't catch this. Does it have to do with this being `Any`?

cirq_superstaq/job.py
```diff
@@ -54,7 +54,7 @@ class Job:
         "data associated with it beyond an id and a status.",
     )
 
-    def __init__(self, client: superstaq_client._SuperstaQClient, job_dict: dict):
+    def __init__(self, client: superstaq_client._SuperstaQClient, job_dict: Dict[str, Any]) -> None:
```
richrines1, line 57: i'm surprised by this one though - I thought we required the use of `-> None` even for `__init__`

vtomole: Does intentionally removing a `-> None` from another `__init__` pass?

richrines1: just found 3 more instances without it, so don't think so..

richrines1: update: mypy does still require `-> None` in the case that `__init__` takes no arguments (i.e. `def __init__(self) -> None:`) because we have the `disallow_untyped_defs` flag on. But if it has any arguments it doesn't qualify as "untyped", and passes because [the return type of `__init__` is implicitly `None`](https://github.com/python/mypy/pull/5677). There doesn't seem to be any way to turn this off

vtomole: Ah.

richrines1:
@vtomole what are your thoughts on using `disallow_any_generics = True` to disallow generic annotations like `x: list` or `-> dict` instead of `x: List[..]` or `-> Dict[.., ..]`?

vtomole (dismissed):
LGTM

vtomole:
@richrines1 Yeah, we should do this.

richrines1:
> @richrines1 Yeah, we should do this.

filed https://github.com/SupertechLabs/applications-superstaq/issues/68 (universal)

singular-value (approved):
+1 lgtm

PR #268, opened 2022-06-15
update qiskit version check

richrines1:
we now require qiskit-terra >= 0.20.0 for ibmq pulse sequences

(also corrects a docstring)

### DIFF
--- a/cirq_superstaq/compiler_output.py
+++ b/cirq_superstaq/compiler_output.py
@@ -4,6 +4,7 @@ from typing import Any, List, Optional, Union
 
 import applications_superstaq
 import cirq
+import packaging.version
 
 import cirq_superstaq as css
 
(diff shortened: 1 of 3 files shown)


singular-value (approved):
Thanks—had been meaning to file an issue for this.

richrines1:
> Can you update https://github.com/SupertechLabs/cirq-superstaq/blob/main/examples/ibmq_compile.ipynb as well?

done. (also set an upper bound to the check in compiler_output)

singular-value:
image
The purple line still needs to be changed right?

richrines1:
> The purple line still needs to be changed right?

yep good call, fixed.

richrines1:
^ oops should've been 0.20.2, will correct once I'm back at my computer

singular-value (approved):
lgtm

PR #269, opened 2022-06-15
Fixed test_cq_compile bug on integration tests, such that it no longe…

SalahedeenIssa:
…r checks whether the circuits are equal, but whether the compiled circuit provides a circuit


vtomole (changes requested):
Revert all the commented out code.

vtomole (approved):

PR #272, opened 2022-06-17
Creating new subfolder ops to simplify construction of custom gates

### LINKED ISSUE #260: Simplify the construction of `Barrier`
Now you need to do

```python
css.Barrier(n).on(*cirq.LineQubit.range(n))
```

Let's create a function
```python
def barrier(*qubits: Tuple[cirq.Qid]) -> cirq.Operation:
    return css.Barrier(len(qubits)).on(*qubits)
```

So people can do 

```python
css.barrier(qubits[0], qubits[1], qubits[2])
```

pgoiporia:
I've added code for simplifying barrier construction, but if we like this structure, we can apply it to other custom gates as well.
Fixes #260


pgoiporia:
Not quite sure how to fix the type check. It's passing locally

pgoiporia:
@singular-value I spoke to @vtomole about the ops folder and he seemed to disagree. How shall we proceed? This branch has commits for both versions

singular-value:
Yeah it's a significant departure from our current design and worth discussing.

Can you message us about it on Tuesday?

pgoiporia:
> Yeah it's a significant departure from our current design and worth discussing.
> 
> Can you message us about it on Tuesday?

Will do

vtomole (changes requested):
Replace everywhere we call  `css.Barrier` with `css.barrier`.

pgoiporia:
> Replace everywhere we call `css.Barrier` with `css.barrier`.

This would be in SuperstaQ though, right? Because I believe barrier is only defined in css, not used in it.

vtomole:
Ah sure, yeah, i'm fine with this being a convenience function.

cirq_superstaq/custom_gates_test.py
```diff
@@ -305,6 +305,11 @@ def test_barrier() -> None:
     assert circuit == cirq.Circuit(operation)
     assert cirq.trace_distance_bound(gate) == 1.0
 
+    # Testing barrier functions
```
vtomole, line 308: Remove comment. What the test doesn't is very clear.

pgoiporia:
> Ah sure, yeah, i'm fine with this being a convenience function.

Will need to merge this in before I can make changes in SuperstaQ

vtomole:
> Creating new subfolder ops to simplify construction of custom gates

Discussed IRL with @singular-value and @pgoiporia. For awareness:  current concensus is to leave things as -s given that the size of `custom_gates.py` is not overwhelming yet and that the only packages that follow this pattern are `cirq-core` and `cirq-google`. 

We've also decided to maintain the flat structure than[ kept things simple](https://en.wikipedia.org/wiki/KISS_principle) in our internal codebase until the amount of code started to become unmaintainable. 

This is something we'll re-visit down the road with the growth of `qss` and `css`.

vtomole (approved):
LGTM

cirq_superstaq/custom_gates.py
```diff
@@ -302,6 +302,10 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> Tuple[str
         return ("|",) * self.num_qubits()
 
 
+def barrier(qubits: Sequence[cirq.Qid]) -> cirq.Operation:
```
richrines1, line 305: @pgoiporia @vtomole out of curiosity why did we decide to make this take a sequence, as opposed to `barrier(*qubits: cirq.Qid)`? it's a bit inconsistent with the rest of `cirq`, e.g.:

```
_ = cirq.measure(q0, q1, q2)  # ok
_ = css.barrier(q0, q1, q2)  # error

_ = cirq.measure([q0, q1, q2])  # error
_ = css.barrier([q0, q1, q2])  # ok
```

vtomole: I overlooked this in review. @pgoiporia Please add a follow up PR that fixes this.

PR #273, opened 2022-06-23
Add resource estimate support

paaige:
Related to: [1386](https://github.com/SupertechLabs/SuperstaQ/pull/1386) and [72](https://github.com/SupertechLabs/applications-superstaq/pull/72)


cirq_superstaq/service.py
```diff
@@ -264,6 +264,19 @@ def get_backends(self) -> dict:
         """Get list of available backends."""
         return self._client.get_backends()["superstaq_backends"]
 
+
+    def resource_estimate(self, circuits: Union[cirq.Circuit, List[cirq.Circuit]], target: str = "neutral_atom_qpu") -> dict:
```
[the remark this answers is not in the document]
paaige: Oh yeah, this should be removed. It was mainly for testing purposes

cirq_superstaq/service.py
```diff
@@ -264,6 +265,35 @@ def get_backends(self) -> dict:
...
+    def resource_estimate(
+        self, circuits: Union[cirq.Circuit, List[cirq.Circuit]], target: str
+    ) -> Union[ResourceEstimate, List[ResourceEstimate]]:
+        """Generates resource estimates for circuit(s).
+
+        Args:
+            circuits: qiskit QuantumCircuit(s).
+            target: string of target representing backend device
+        Returns:
+            ResourceEstimate object containing gate count, critical path length, and error estimate.
```
singular-value, line 277: us
```suggestion
            ResourceEstimate(s) containing resource costs (after compilation)
```

cirq_superstaq/service.py
```diff
@@ -264,6 +265,35 @@ def get_backends(self) -> dict:
...
+        serialized_circuit = css.serialization.serialize_circuits(circuits)
+
+        request_json = {
+            "cirq_circuits": serialized_circuit,
+            "backend": target,
+        }
+
+        json_dict = self._client.resource_estimate(request_json)
+
+        # Type must be specified due to mypy issue: https://github.com/python/mypy/issues/6898
```
singular-value, line 289: you can use same return code as in IBM

cirq_superstaq/service.py
```diff
@@ -264,6 +265,36 @@ def get_backends(self) -> dict:
...
+
+        request_json = {
+            "cirq_circuits": serialized_circuit,
+            "backend": target,
+        }
+
+        json_dict = self._client.resource_estimate(request_json)
+
+        resource_estimates = [
+            ResourceEstimate(json_data=resource_estimate)
```
vtomole, line 290: Not relevant to this PR, but why does `ResourceEstimate` allow one to pass in json?

vtomole (dismissed):
LGTM

cirq_superstaq/service.py
paaige: We discussed having to/from json conversion methods in the ResourceEstimate class but because ResourceEstimate is a dataclass there's a built-in method that runs after the default constructor (PostInit) and initializes the object based on any initialization variables specified. It seemed simpler to just be able to construct a ResourceEstimate object from a dictionary rather than having another method (outside of the data class) that does this.

vtomole: :+1:

vtomole (approved):
LGTM

PR #278, opened 2022-06-28
Metadata Inclusion

### LINKED ISSUE #277: Include metadata for parallel gates
We want to use parallel gates from css in order to schedule. This way, we can pack gates that are already scheduled in the same moment as a set of parallel gates, such that they cannot be broken up later on in another iteration of the algorithm. In order to do this, we must include metadata for the class of parallel gates.

SalahedeenIssa:
Fixes #277


dowusu-antwi:
(No longer needed, there's a better way to do it)

PR #287, opened 2022-07-05
Replace running on bogota to running on jakarta

vtomole:


cirq_superstaq/daily_integration_test.py
```diff
@@ -23,7 +23,7 @@ def test_ibmq_compile(service: css.Service) -> None:
     out = service.ibmq_compile(circuit, target="ibmq_jakarta_qpu")
     assert isinstance(out.circuit, cirq.Circuit)
     assert out.pulse_sequence is not None
-    assert 800 <= out.pulse_sequence.duration <= 1000  # 896 as of 12/27/2021
+    assert 800 <= out.pulse_sequence.duration <= 1000  # 3616 as of 6/30/2022
```
singular-value, line 26: ?

singular-value (changes requested):
L26 is weird

cirq_superstaq/daily_integration_test.py
vtomole: What is it?

singular-value (changes requested):
Does css also need upgraded applications_ss version?

We currently have a >= requirement which means that the integration test presumably uses the latest applications_ss version, but it doesn't necessarily work with lower version.

I think we probably want to pin the applications_ss version to a specific latest version, as you've done in qss then?

cirq_superstaq/daily_integration_test.py
vtomole: Ah, it was the same code. reverted.

pgoiporia (approved):
LGTM but why are we pinning applications_superstaq?

vtomole:
Jumping on top of @pgoiporia's question, the previous line `applications-superstaq>=0.1.5` will get the latest version of `app_ss`. But the new commit is just for qss-css parity. If we do `applications-superstaq>=0.1.5` here we need to do `applications-superstaq>=0.1.5` in qss.

singular-value (approved):

PR #289, opened 2022-07-06
upper bound python-rapidjson

richrines1:
otherwise macos check fails due to https://github.com/python-rapidjson/python-rapidjson/issues/166

should be reverted once `python-rapidjson` has been patched


vtomole (approved):
LGTM

PR #295, opened 2022-07-08
Add resource estimate example / jupyter notebook

paaige:


pgoiporia (approved):
LGTM

vtomole:
@paaige You'll need to do the same thing for qss as well.

PR #298, opened 2022-07-12
Changing barrier qubit input from list to unpacked tuple

pgoiporia:
This makes the `css.barrier()` call symmetric to `cirq.measure()`


richrines1 (approved):
lgtm!

PR #301, opened 2022-07-13
Make css.Barrier a cirq.InterchangeableQubitsGate

richrines1:
so that e.g. `css.barrier(q0, q1, q2) == css.barrier(q2, q1, q0)`


vtomole (approved):
Good improvement. thx.

vtomole (approved):

PR #310, opened 2022-07-22
Use fuzz for testing jaqal string equivalency

vtomole:


vtomole:
This passes: https://github.com/SupertechLabs/cirq-superstaq/actions/runs/2721737206

requirements.txt
```diff
@@ -2,3 +2,4 @@ applications-superstaq==0.1.15
 cirq>=0.14.0
 qubovert>=1.2.3
 sympy<1.10
+thefuzz[speedup]~=0.19.0
```
richrines1, line 5: can this go in dev-requirements?

cirq_superstaq/daily_integration_test.py
```diff
@@ -121,7 +122,8 @@ def test_qscout_compile(service: css.Service) -> None:
     cirq.testing.assert_circuits_with_terminal_measurements_are_equivalent(
         out.circuit, compiled_circuit, atol=1e-08
     )
-    assert out.jaqal_program == jaqal_program
+    ratio = fuzz.ratio(out.jaqal_program, jaqal_program)
+    assert ratio >= 85
```
richrines1, line 126: would a higher threshold work here?

richrines1: the circuit above probably needs to be regenerated to include the import statement/new cirq version/etc

vtomole: > would a higher threshold work here?

Yes, I could copy-paste the new output and increase the threshold but that might defeat the purpose of using `fuzz` here in the first place: We want `almost_equivalent` programs; basically this test should only failed if the returned program is  super.wrong.

vtomole: Sorry, to answer your question explicitly: doing `>=89` without changing anything else will pass this but i felt that was a "weird" random value compared to 85 sounded nice cause it's a power of 5.

richrines1: sure, but right now the string we're comparing against is super different from any expected result (and isn't even a working jaqal program) due to the lack of import statement

what would it mean for this endpoint to be broken? it seems like maybe the most important thing to check in an integration test is just that it returns a valid jaqal program, which we could do with e.g.
``` 
_ = jaqalpaq.parser.parse_jaqal_string(jaqal_str)
```
instead of trying to compare the actual string to something specific

vtomole: I've asserted that only a part of the jaqal string (measurement) exists  to be consistent to what we do below..

richrines1 (approved):
lgtm

vtomole:
Confirmed that it passes: https://github.com/SupertechLabs/cirq-superstaq/runs/7506231827?check_suite_focus=true

PR #312, opened 2022-07-28
Integrate gpu simulator as service

paaige:


cirq_superstaq/service.py
```diff
@@ -263,6 +264,22 @@ def get_backends(self) -> dict:
...
+    ) -> str:
+        serialized_circuit = css.serialization.serialize_circuits(circuit)
+
+        request_json =  {
+            "cirq_circuits": serialized_circuit,
+            "backend": target,
+            "gpu": gpu,
+        }
+
+        amplitudes = self._client.simulate(request_json)
```
vtomole, line 278: Open a draft PR in gss for this method.

cirq_superstaq/service.py
```diff
@@ -263,6 +264,22 @@ def get_backends(self) -> dict:
         """Get list of available backends."""
         return self._client.get_backends()["superstaq_backends"]
 
+    def simulate(
```
vtomole, line 267: We've decided to get rid of this function in place of passing in `method` and `options` arguments to `create_job`

vtomole (changes requested):

cirq_superstaq/service.py
paaige: that's from a very old commit

cirq_superstaq/service.py
```diff
@@ -229,7 +233,9 @@ def create_job(
             serialized_circuits={"cirq_circuits": serialized_circuits},
             repetitions=repetitions,
             target=target,
+            method=method,
             ibmq_pulse=ibmq_pulse,
+            options={"options": json.dumps(options)},
```
vtomole, line 238: Link

vtomole (commented):
Need to pass checks.

vtomole (changes requested):
Add tests for this `method` and `serialized_options` arguments.

vtomole:
After that this should be good to go.

vtomole (approved):
LGTM

vtomole (approved):
LGTM!

PR #317, opened 2022-08-09
new aqt demo

singular-value:
from notebook @richrines1 made


vtomole (approved):
Good to merge after nit.

vtomole (approved):
LGTM

vtomole (approved):
LGTM

requirements.txt
```diff
@@ -1,4 +1,4 @@
-cirq>=0.14.0
-general-superstaq~=0.2.4
+cirq~=1.0.0
```
richrines1, line 1: @singular-value why was this changed? do we want to drop support for cirq==0.14.*?

vtomole: Didn't catch this but @singular-value might have seen https://github.com/SupertechLabs/general-superstaq/issues/130 hence it's misleading-ness here.

vtomole: I'm fine with letting this through. SupermarQ is on cirq-1.0 now.

PR #319, opened 2022-08-11
adding swap mirroring option to qscout_compile with tests

AndrewLitteken:
Discussed previously, swap mirroring is not the most common, and while it's good to have on by default, it may not be the expectation, and a user may have reasons to turn it off (explicit testing, or measurement outcomes).  This adds an option to cirq-superstaq to make it user configurable.

PR #325, opened 2022-08-23
updating requirements

AndrewLitteken:


requirements.txt
```diff
@@ -1,4 +1,4 @@
-cirq>=0.14.0
-general-superstaq~=0.2.1
+cirq~=1.0.0
```
richrines1, line 1: do we need cirq 1.0.0? usually we try to support a version or two back

AndrewLitteken: Maybe not, but it's what the checks were requiring

richrines1: ah that sounds like a bug in the check script - filed https://github.com/SupertechLabs/general-superstaq/issues/130

(i think it should still pass without the change though - i believe the cirq/qiskit version checks are just warnings)

richrines1 (approved):
lg (though i think this would also be closed by merging #319)

AndrewLitteken:
True, but probably a better idea to keep unrelated fixes separate

PR #331, opened 2022-09-02
Generalize `_version_test`

vtomole:
This makes it so you don't need to update version test when updating the version.


richrines1 (approved):
lgtm. assuming this will be replicated in the other repos?

vtomole:
> assuming this will be replicated in the other repos?


Very weird that `import packaging` doesn't work in other repos. I have to do `import packaging.version`. Will look into it this evening.

richrines1:
> Very weird that `import packaging` doesn't work in other repos. I have to do `import packaging.version`. Will look into it this evening.

`packaging.__init__` doesn't import any submodules itself, so `packaging.version` should always need to be explicitly imported somewhere. i think it works in css because `cirq` imports `matplotlib`, and `matplotlib.__init__` already imports stuff from `packaging.version`. (imo for consistency we should use `import packaging.version` here too though)

stephanielee9 (approved):
lgtm - should we eventually implement a version test in supermarq as well?

vtomole:
> should we eventually implement a version test in supermarq as well?

Yes. Done: https://github.com/SupertechLabs/SupermarQ/pull/48

PR #333, opened 2022-09-06
fix `default_target`

richrines1:
`default_target` should now work as advertised

removes any dependency on `gss.SuperstaqClient`'s `default_target` argument, so https://github.com/SupertechLabs/general-superstaq/issues/142 can be tackled with without any conflicts after merging this and https://github.com/SupertechLabs/qiskit-superstaq/pull/222


vtomole:
Don't we want to get rid of `default-target` everywhere? https://github.com/SupertechLabs/general-superstaq/issues/142#issuecomment-1238415312

richrines1:
> Don't we want to get rid of `default_target` everywhere? [SupertechLabs/general-superstaq#142 (comment)](https://github.com/SupertechLabs/general-superstaq/issues/142#issuecomment-1238415312)

ah maybe. i was thinking it still might be a little nice here just to avoid writing `target=..` everywhere (as in the code example in https://github.com/SupertechLabs/general-superstaq/issues/142#issue-1363522555), but wasn't really sure. happy to remove if you think that doesn't make sense/isn't super useful

vtomole:
It is a little useful but qss doesn't have it so i'm leaning on parity here.

richrines1:
> It is a little useful but qss doesn't have it so i'm leaning on parity here.

oh i was kinda thinking this was better parity, in that in qss you kinda do this when you instantiate a backend. E.g.
```
service = css.Service(default_backend="ibmq_qasm_simulator")

job1 = service.create_job(circuit1, shots)
job2 = service.create_job(circuit2, shots)
...
```
seems more similar to
```
provider = qss.SuperstaQProvider()
backend = provider.get_backend("ibmq_qasm_simulator")

job1 = backend.run(circuit1, shots)
job2 = backend.run(circuit2, shots)
...
```

vtomole:
Ah, you're right.

PR #338, opened 2022-09-21
use disallow-any-generics mypy flag

vtomole:
Part of https://github.com/SupertechLabs/general-superstaq/issues/68


SalahedeenIssa (approved):
LGTM

cirq_superstaq/custom_gates.py
```diff
@@ -31,7 +33,7 @@ class ZZSwapGate(cirq.Gate, cirq.ops.gate_features.InterchangeableQubitsGate):
     For :math:`\theta = 0`, the ZZ-SWAP gate is just an ordinary SWAP.
     """
 
-    def __init__(self, theta: cirq.TParamVal) -> None:
+    def __init__(self, theta: Optional[float] = None) -> None:
```
richrines1, line 36: why this change? I made this gate work with sympy expressions

cirq_superstaq/custom_gates.py
```diff
@@ -445,7 +447,7 @@ def phi(self) -> float:
     def theta(self) -> float:
         return self.exponent * np.pi
 
-    def __pow__(self, power: float) -> "RGate":
+    def __pow__(self, power: Union[float, sympy.Expr]) -> "RGate":
```
richrines1, line 450: fine as is, but fwiw this is equivalent to
```suggestion
    def __pow__(self, power: cirq.TParamVal) -> "RGate":
```
(elsewhere as well)

vtomole: Done.

cirq_superstaq/custom_gates.py
```diff
@@ -445,7 +446,7 @@ def phi(self) -> float:
     def theta(self) -> float:
         return self.exponent * np.pi
 
-    def __pow__(self, power: float) -> "RGate":
+    def __pow__(self, power: cirq.TParamVal) -> "RGate":
```
richrines1, line 449: i think this one should be reverted - `RGate.__init__` only accepts floats (for now)
```suggestion
    def __pow__(self, power: float) -> "RGate":
```

cirq_superstaq/job_test.py
```diff
@@ -31,8 +32,13 @@ def new_job() -> css.Job:
     return css.Job(client, "my_id")
 
 
-def mocked_get_job_requests(*job_dicts: Dict[str, Any]) -> mock._patch:
-    """Mocks the server's response to `get_job` requests using the given sequence of job_dicts."""
+def mocked_get_job_requests(
+    *job_dicts: Dict[str, Any]
+) -> "mock._patch[Union[MagicMock, AsyncMock]]":
```
richrines1, line 37: nit
```suggestion
) -> "mock._patch[Union[mock.MagicMock, mock.AsyncMock]]":
```
(we use "mock.MagicMock" elsewhere in this file so it feels weird to import it separately for this line)

richrines1: also fwiw - not sure which is more correct in this case, but it seems to pass type check for me with either
```suggestion
) -> "mock._patch[mock.MagicMock]":
```
or
```suggestion
) -> "mock._patch[mock.Mock]":
```

vtomole: Will do the last suggestion.

cirq_superstaq/compiler_output.py
```diff
@@ -32,7 +32,7 @@ def __init__(
         pulse_sequences: Optional[Any] = None,
         seq: Optional["qtrl.sequencer.Sequence"] = None,
         jaqal_programs: Optional[Union[List[str], str]] = None,
-        pulse_lists: Optional[Union[List[List], List[List[List]]]] = None,
+        pulse_lists: Optional[Union[List[List[str]], List[List[List[str]]]]] = None,
```
richrines1, line 35:
```suggestion
        pulse_lists: Optional[Union[List[List[List[Any]]], List[List[List[List[Any]]]]]] = None,
```

https://github.com/SupertechLabs/qiskit-superstaq/pull/234#discussion_r1006241901

cirq_superstaq/custom_gates.py
richrines1: should this have been resolved? it seems like either this signature or `RGate.__init__`'s is wrong

SalahedeenIssa (dismissed):

cirq_superstaq/custom_gates.py
vtomole: Applied your suggestion. Mypy still passes.

richrines1 (approved):
lgtm!

PR #347, opened 2022-10-03
fix circuit diagram for ParallelGates

richrines1:
updates to `ParallelGatesDiagram._circuit_diagram_info_`:
* fall back to default cirq diagram if any component gate doesn't implement `_circuit_diagram_info_` (fixes: #346)
* construct `CircuitDiagramInfoArgs` correctly for each component gate, instead of just passing the input `args` everywhere (modeled on `cirq.ControlledGate`)


perlinm (approved):
I'll admit there's a bit of magic going on that I don't understand involving `cirq.CircuitDiagramInfoArgs` and `cirq.circuit_diagram_info`, but if it works then it's nbd, especially since this is just about pretty printing

cirq_superstaq/custom_gates.py
```diff
@@ -385,22 +385,39 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
...
-            diagram_info = cirq.circuit_diagram_info(gate, args)
-            full_wire_symbols = diagram_info._wire_symbols_including_formatted_exponent(
-                args,
-                preferred_exponent_index=cirq.num_qubits(gate) - 1,
+            qubit_index = len(wire_symbols_with_subscripts)
+            num_qubits = gate.num_qubits()
+            sub_args = cirq.CircuitDiagramInfoArgs(
+                known_qubit_count=(num_qubits if args.known_qubit_count is not None else None),
+                known_qubits=(
+                    args.known_qubits[qubit_index:][:num_qubits]
```
perlinm, line 395: Actually hold up, shouldn't `args.known_qubits` [simply be an `Iterable` here](https://quantumai.google/reference/python/cirq/CircuitDiagramInfoArgs)?  If so this does not make sense (and if not, then why?)

richrines1:
most of the magic was just copied over from `cirq.ControlledGate._circuit_diagram_info_` (which similarly incorporates the diagram of an arbitrary sub_gate)

PR #348, opened 2022-10-03
parallel_gates_operation

richrines1:
adds a `css.parallel_gates_operation()` helper function, such that
```
css.parallel_gates_operation(op1, op2, ...)
```
is equivalent to
```
css.ParallelGates(op1.gate, op2.gate, ...).on(*op1.qubits, *op2.qubits, ...)
```
with appropriate input validation


singular-value (approved):
looks great!

PR #349, opened 2022-10-04
ignore flake8 import order

richrines1:
prevents flake8 from failing if `flake8-import-order` happens to be installed (since https://github.com/SupertechLabs/general-superstaq/pull/134 we enforce a different import order via `isort`)

(parity w/ https://github.com/SupertechLabs/general-superstaq/pull/166)


knsmith (approved):
Confirmed that ignore-flake8-import-order prevents error.

setup.cfg
```diff
@@ -23,6 +23,7 @@ max-complexity = 10
 max-line-length = 100
 show_source = True
 statistics = True
+extend-ignore = I100, I101, I201, I202
```
perlinm, line 26: For the sake of clarity/self-documentation, how about something like this?

```suggestion
extend-ignore = I100, I101, I201, I202  # ignore errors thrown by flake8-import-order (if it is installed), which is superceded by isort
```

perlinm (approved):
lgtm

PR #366, opened 2022-10-21
put eca flags in options dict

richrines1:
(superstaq already looks for them there, so currently they're just being ignored)

also adds an integration test to make sure the `num_equivalent_circuits` flag is actually respected now


paaige (approved):
Lgtm.

PR #367, opened 2022-10-24
Issue form templates

stephanielee9:


stephanielee9:
Creating yaml files doesn't seem to do the trick (from results of https://github.com/SupertechLabs/qiskit-superstaq/pull/252 and https://github.com/SupertechLabs/general-superstaq/pull/183) - seems like we need to go through https://github.com/SupertechLabs/cirq-superstaq/issues/templates/edit instead

hallben4 (approved):
lgtm

PR #376, opened 2022-11-07
convert `__eq__` to `cirq.value_equality`

### LINKED ISSUE #356: Replace all `__eq__` methods with `cirq.value_equality`
Cleaner than defining `__eq__` for every class.

dowusu-antwi:
Fixes #356.

In same spirit as server-side changes currently underway. This allows us to more simply define equality tests for a given class / object: return which values to compare, rather than e.g. having to do a check like `isinstance(other, type(self))`, etc.

Also expands usage of `job` fixture to other tests in `job_test.py` (except `test_job_status_refresh` b/c fixtures can't be reinitialized inside a test, AFAIK).


cirq_superstaq/job.py
```diff
@@ -170,7 +172,5 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"css.Job(client={self._client!r}, job_id={self.job_id()!r})"
 
-    def __eq__(self, other: Any) -> bool:
-        if not isinstance(self, type(other)):
-            return False
-        return self._job == other._job
+    def _value_equality_values_(self) -> Any:
```
richrines1, line 175:
```suggestion
    def _value_equality_values_(self) -> Dict[str, Any]:
```

richrines1 (commented):
lgtm pending nit

richrines1 (approved):

PR #350, opened 2022-10-05
global phase handling in custom gates

richrines1:
corrects handling of global phase in equality comparisons:
* strict equality (i.e. `gate1 == gate2`) and approximate equality (`cirq.approx_eq(gate1, gate2)`) are now True only when gates have the same global phase
* `cirq.equal_up_to_global_phase()` is supported correctly by all gates
* fixes global phase discrepancy between `css.AceCR("+-", x)` and `qss.AceCR("+-", x)` for some `x`
* gates now evaluate as equivalent to appropriate parent classes, so for example the following are now all True:
```
css.RGate(pi, pi / 2) == cirq.PhasedXPowGate(phase_exponent=0.5, global_shift=-0.5)
cirq.approx_eq(css.RGate(pi, pi / 4), cirq.PhasedXPowGate(exponent=1 + 1e-12, phase_exponent=0.25 + 1e-12, global_shift=-0.5))
cirq.equal_up_to_global_phase(css.RGate(pi, pi / 2), cirq.PhasedXPowGate(phase_exponent=0.5))
css.ParallelRGate(x, y, n) == cirq.ParallelGate(css.RGate(x, y), n)
```

fixes: #230
fixes: #250

PR #383, opened 2022-11-11
approximate value comparison for ParallelRGate

richrines1:
the `@value_equality` decorator was removed from `ParallelRGate` in #350 so that e.g. `cirq.ParallelGate(css.RGate(1, 2), 3) == cirq.ParallelRGate(1, 2, 3)`. Unfortunately this came at the expense of approximate comparison. The latter being much more important, this reverts that change.

summary:
```
# fails now, will pass after this pr:
assert cirq.approx_eq(css.ParallelRGate(1, 2, 3), css.ParallelRGate(1 + 1e-10, 2 + 1e-10, 3))

# passes now, will fail after this pr:
assert css.ParallelRGate(1, 2, 3) == cirq.ParallelGate(css.RGate(1, 2), 3)
assert cirq.approx_eq(css.ParallelRGate(1, 2, 3), cirq.ParallelGate(css.RGate(1, 2), 3))

# passes before and after this pr:
assert cirq.equal_up_to_global_phase(css.ParallelRGate(1, 2, 3), cirq.ParallelGate(css.RGate(1, 2), 3))
```


pgoiporia (approved):
LGTM!

PR #384, opened 2022-11-15
Implement `statevector` function

vtomole:


stephanielee9:
To be continued by Victory later in 2023

PR #399, opened 2022-12-14
Add warning about QTRL wheel

### LINKED ISSUE #77: `aqt.ipynb` should have default configs
In https://github.com/SupertechLabs/cirq-superstaq/blob/main/examples/aqt_compile.ipynb, there are no default yaml configs, so if user doesn't provide configs there's an error:

no-configs-aqt-error

dowusu-antwi:
Fixes #77. Assuming we still don't want to store configs in the repo, the errors (and new warning) are descriptive enough.
image


cirq_superstaq/compiler_output.py
```diff
@@ -51,6 +51,11 @@ def __init__(
             self.pulse_sequences = pulse_sequences
             self.jaqal_programs = jaqal_programs
 
+        if seq is None:
```
richrines1, line 54: `CompilerOutput` is also used by the other compile endpoints, so this warning should be moved to `read_json_aqt()`. there i think it can be handled by adding an `else` to the `if importlib.util.find_spec..` block. (see the similar warning for pulser in `read_json_neutral_atom()`)

richrines1 (commented):
this is definitely a good addition but i'm not sure it fully solves #77 --- i can have qtrl installed but still not have uploaded configs

cirq_superstaq/compiler_output.py
richrines1: > As for the configs, how about we keep default copies of `KeysightPulses` and `KeysightVariables` in the repo?

my understanding is that we don't want to make them publicly available

cirq_superstaq/compiler_output.py
```diff
@@ -155,6 +155,11 @@ def _sequencer_from_state(state: Dict[str, Any]) -> "qtrl.sequencer.Sequence":
             state["_readout"] = readout_seq
 
         seq = _sequencer_from_state(state)
+    else:
+        warnings.warn(
+            "Your sequence for this output is None. Please make sure you have the QTRL wheel "
```
richrines1, line 160: nit
```suggestion
            "Your sequence for this output is None. Please make sure you have the qtrl package "
```

cirq_superstaq/compiler_output.py
```diff
@@ -155,6 +155,11 @@ def _sequencer_from_state(state: Dict[str, Any]) -> "qtrl.sequencer.Sequence":
             state["_readout"] = readout_seq
 
         seq = _sequencer_from_state(state)
+    else:
+        warnings.warn(
+            "Your sequence for this output is None. Please make sure you have the qtrl package "
+            "installed."
```
richrines1, line 161: ah sorry one more nit
```suggestion
            "installed in order to de serialize compiled pulse sequences."
```

richrines1:
actually can we add a `pytest.warns` check to `compiler_output_test.test_read_json_aqt()`?

dowusu-antwi:
> actually can we add a `pytest.warns` check to `compiler_output_test.test_read_json_aqt()`?

Done.

setup.cfg
```diff
@@ -28,7 +28,6 @@ extend-ignore = E203, I100, I101, I201, I202  # these errors conflict with black
 [mypy]
 ignore_missing_imports = True
 show_error_codes = True
-disallow_any_generics = True
```
richrines1, line 31: this doesn't need to be changed. https://github.com/SupertechLabs/general-superstaq/pull/201 will fix the check/setup.py error (the check only happens locally and should just be warning)

richrines1 (dismissed):
lgtm after reverting setup.cfg

setup.cfg
dowusu-antwi: Right! I keep forgetting.

dowusu-antwi: Done, please re-review -- thanks!

(Too bad `git revert` undoes approval, although technically better practice than a new commit I suppose.)

richrines1 (approved):
(still) lgtm :)

we should also make the same changes in qss

PR #404, opened 2022-12-28
Regression test for setting random seed for ECA compiling in AQT

singular-value:


singular-value:
See also https://github.com/SupertechLabs/qiskit-superstaq/pull/289

richrines1 (approved):
lgtm!

richrines1:
(confirmed it passes locally)

PR #405, opened 2023-01-06
Caching Github Actions

pgoiporia:


pgoiporia:
About 2x speedup

vtomole (approved):
Yep, this is noticeable in the Windows check (10 mins -> 3 mins). Very kewl!

vtomole (changes requested):
Look at how Cirq caches the Windows checks.

vtomole (changes requested):
Add windows tests

vtomole (approved):
LGTM

vtomole (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -22,6 +22,10 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.8
+      - uses: actions/cache@v2
+        with:
+          path: ${{ env.pythonLocation }}
+          key: ${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ hashFiles('dev-requirements.txt') }}
```
richrines1, line 28: @pgoiporia @vtomole @dowusu-antwi i just noticed that we're not including `requirements.txt` here - should we be? /could this be causing some of the weird behavior we noticed before?

dowusu-antwi: > i just noticed that we're not including `requirements.txt` here - should we be?

I think so, but I'll let Palash weigh in to confim.

> could this be causing some of the weird behavior we noticed before?

This could be related, but it was a server-side issue and we include `requirements.txt` in the SuperstaQ `yaml`.

richrines1: fixed in #1358 - it looks like hashFiles accepts multiple files/wildcards, so we can just use `*requirements.txt` to make sure we're including all requirements

PR #407, opened 2023-01-10
Update qiskit pins

vtomole:


richrines1 (approved):
lgtm

vtomole:
Notebook test will continue to fail until qiskit is updated in the server. Merging.

PR #408, opened 2023-01-10
helper to returns the indices of all measured qubits

richrines1:
(requested by AQT)


paaige (approved):

cirq_superstaq/compiler_output_test.py
```diff
@@ -26,6 +26,31 @@ def test_active_qubit_indices() -> None:
...
+        cirq.measure(q4, q1),
+        cirq.measure(q6),
+    )
+    assert css.measured_qubit_indices(circuit) == [1, 3, 4, 6]
+
+    subcircuit = cirq.Circuit(cirq.X(q0), cirq.measure(q1))
+    subcircuit += cirq.CircuitOperation(subcircuit.freeze(), qubit_map={q0: q3, q1: q2})
+
+    circuit = cirq.Circuit(
+        cirq.CircuitOperation(subcircuit.freeze()).with_qubits(q5, q2, q4, q1),
```
SalahedeenIssa, line 45: How/why is with_qubits used here? And what is the final mapping of this circuit? The q5 is confusing me a lot.

richrines1: cleaned up/better documented the test, and replaced `with_qubits` with `with_qubit_mapping` to hopefully make it more clear. (the idea is just to make sure that the qubit map of CircuitOperations are reflected in the measured qubit indices)

(also realized i could lean on cirq's `unroll_circuit_op` to replace most of the logic in `measured_qubit_indices`, which makes having a complicated test case here less important)

cirq_superstaq/compiler_output_test.py
```diff
@@ -26,6 +26,38 @@ def test_active_qubit_indices() -> None:
...
+    # Check that measurements in CircuitOperations are mapped correctly
+    circuit = cirq.Circuit(cirq.measure(q1))
+    subcircuit = cirq.FrozenCircuit(cirq.X(q0), cirq.measure(q2, q3))
+
+    assert css.measured_qubit_indices(circuit) == [1]
+    assert css.measured_qubit_indices(subcircuit) == [2, 3]
+    assert css.measured_qubit_indices(circuit + subcircuit) == [1, 2, 3]
+
+    # Apply subcircuit to qubits [q6, q5, q3] of circuit (measurements should land on q3 and q5)
+    circuit += cirq.CircuitOperation(subcircuit).with_qubit_mapping({q1: q6, q2: q5, q3: q3})
```
SalahedeenIssa, line 50: Just for my own understanding - since in the initial circuit  + subcircuit the indices measured are 1,2,3 , and then 1 is mapped onto 6, 2 is mapped onto 5 and 3 is mapped onto 3, shouldn't 6 be measured along with 5 and 3 and not 1?

richrines1: the mapping only applies to the circuit operation, so `subcircuit`'s measurements on (q2, q3) are mapped to (q5, q3) and `circuit`'s measurement on q1 stays where it is. i just refactored again to emphasize this, lmk if you think it's more clear

(thank you for this - these tests are supposed to double as documentation so if they're not clear they should be changed)

SalahedeenIssa: Ahh I see, that makes a lot of sense now. Thank you!

cirq_superstaq/compiler_output_test.py
```diff
@@ -38,20 +38,31 @@ def test_measured_qubit_indices() -> None:
...
 
-    # Check that measurements in CircuitOperations are mapped correctly
-    circuit = cirq.Circuit(cirq.measure(q1))
-    subcircuit = cirq.FrozenCircuit(cirq.X(q0), cirq.measure(q2, q3))
 
-    assert css.measured_qubit_indices(circuit) == [1]
+def test_measured_qubit_indices_with_circuit_operations() -> None:
+    """Check that measurements in CircuitOperations are mapped correctly."""
+
+    q0, q1, q2, q3, _, q5, q6 = cirq.LineQubit.range(7)
```
SalahedeenIssa, line 45: Why no q4? Could you just use 6 qubits?

cirq_superstaq/compiler_output_test.py
```diff
@@ -38,20 +38,31 @@ def test_measured_qubit_indices() -> None:
...
+    subcircuit_op_mapped = cirq.CircuitOperation(subcircuit).with_qubit_mapping({q1: q6, q2: q5})
+    assert css.measured_qubit_indices(subcircuit_op_mapped.mapped_circuit()) == [3, 5]
+
+    # Check that measured_qubit_indices() respects the qubit mapping
+    circuit = cirq.Circuit(cirq.measure(q1))
+    assert css.measured_qubit_indices(circuit) == [1]
+    assert css.measured_qubit_indices(circuit + subcircuit_op_no_map) == [1, 2, 3]  # no mapping
+    assert css.measured_qubit_indices(circuit + subcircuit_op_mapped) == [1, 3, 5]  # with mapping
 
     # Double-check against the unrolled circuit
```
SalahedeenIssa, line 64: Maybe make this comment a little more detailed to include the fact that you are only checking to see if the unrolling alters the mapping of the qubits and their measurements.

richrines1: done

cirq_superstaq/compiler_output_test.py
richrines1: basically just to demonstrate that "qubit indices" refers to `qubit.x` (so e.g. `cirq.LineQubit(6)` has index 6), regardless of the actual number of qubits in the circuit. Added comments and also updated the docstrings of `measured_qubit_indices` and `active_qubit_indices` to say as much

SalahedeenIssa: Ah okay this makes sense.

SalahedeenIssa (approved):
lgtm!

PR #409, opened 2023-01-11
Update `version.py` to `0.3.12`

vtomole:


dowusu-antwi (approved):

cirq_superstaq/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.3.10"
+__version__ = "0.3.12"
```
dowusu-antwi, line 1: (I should've caught this -- it's a good lesson in closely reviewing even minor changes.)

PR #413, opened 2023-01-17
Create custom stripped CZ gate

SalahedeenIssa:
Fixes https://github.com/SupertechLabs/cirq-superstaq/issues/412


perlinm:
fixes #412

cirq_superstaq/custom_gates.py
```diff
@@ -572,6 +572,28 @@ def _json_dict_(self) -> Dict[str, Any]:
...
+
+    def _num_qubits_(self) -> int:
+        return 2
+
+    def _unitary_(self) -> Optional[npt.NDArray[np.complex_]]:
+        return np.array(
+            [
+                [1.0, 0.0, 0.0, 0.0],
+                [0.0, 1.0, 0.0, 0.0],
+                [0.0, 0.0, 1.0, 0.0],
```
perlinm, line 588: I think these should be phases too
```suggestion
                [0.0, np.exp(1j * self.phi), 0.0, 0.0],
                [0.0, 0.0, np.exp(1j * self.phi), 0.0],
```

cirq_superstaq/custom_gates.py
```diff
@@ -572,6 +572,28 @@ def _json_dict_(self) -> Dict[str, Any]:
         return cirq.protocols.obj_to_dict_helper(self, ["theta", "phi"])
 
 
+class StrippedCZGate(cirq.Gate):
+    def __init__(self, phi=0) -> None:
+        super(StrippedCZGate, self)
```
perlinm, line 577: What does this line do?

SalahedeenIssa: I wanted to stay as close to the format shown in this [https://quantumai.google/cirq/build/custom_gates](link) as possible to start off, and then make changes as necessary.

cirq_superstaq/custom_gates.py
SalahedeenIssa: Ah right good catch.

SalahedeenIssa:
New branch made instead

perlinm:
note: new branch is #441

PR #420, opened 2023-01-20
make bswap swap 11<->22 levels instead of 02<->11

### LINKED ISSUE #419: BSWAP should be between (2,2) and (1,1)
https://github.com/SupertechLabs/cirq-superstaq/blob/898d9d183da6f35f0d07d4f89207a578d53bffcf/cirq_superstaq/ops/qudit_gates.py#L21

richrines1:
fixes #419


singular-value (approved):
Nice on keeping the code general so that this was a easy change

PR #421, opened 2023-01-23
Parametric Cross-Resonance

dowusu-antwi:
Towards fixing server-side issue, augments `AceCR` to take a `theta` parameter (halved for echo implementation) (requires server-side changes for compilation):

image


cirq_superstaq/ops/qubit_gates.py
```diff
@@ -230,14 +230,18 @@ class AceCR(cirq.Gate):
...
     Args:
         polarity: Should be either "+-" or "-+". Specifies if positive or negative half-CR is first
+        theta: Angle of rotation for CR gate (i.e., twice the angle for each echoed half-CR).
         sandwich_rx_rads: Angle of rotation for an rx gate applied to the "X side" simultaneously
             with the X gate on the "Z side".
     """
 
-    def __init__(self, polarity: str, sandwich_rx_rads: cirq.TParamVal = 0) -> None:
+    def __init__(
+        self, polarity: str, theta: float = np.pi / 2, sandwich_rx_rads: cirq.TParamVal = 0
```
richrines1, line 239: can we do away with the `polarity` argument now?

dowusu-antwi: Sounds good to me.

(I forgot to post this comment on the server-side PR: in support of `one issue <-> one PR`, should that be a separate batch of PRs? Although, it's also a breaking change -- which supports including it now.)

richrines1: hmm i guess my instinct is to consider "replace polarity string with parameter" to be a single issue, but if it seems a lot simpler to break up that's definitely reasonable too

btw this might be totally unnecessary but it occurs to me we could also use something like this if we wanted to maintain backwards compatibility:
```python
def __init__(self, theta: Union[str, cirq.TParamVal] = np.pi / 2, sandwich_rx_rads: cirq.TParamVal = 0):
    if theta == "+-":
        theta = np.pi / 2
    elif theta == "-+":
        theta = -np.pi / 2
    elif isinstance(theta, str):
        raise ValueError(...)
```

dowusu-antwi: Done. (But now there's an unfortunate deadlock -- notebook checks won't pass without server-side `polarity` parameter change, but server-side changes can't be merged without `css.AceCR` from here...)

dowusu-antwi: Added temporary fix: commented out relevant notebook lines in `ibmq_compile.ipynb` and `resource_estimate.ipynb` -- will uncomment in new PR after deploying this and server-side changes.

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -246,71 +254,111 @@ def _num_qubits_(self) -> int:
...
+            sandwich_rx_rads = self.sandwich_rx_rads % (4 * np.pi)
 
-        return self.polarity, self.sandwich_rx_rads % (4 * np.pi)
+        return theta, sandwich_rx_rads
 
-    def _value_equality_approximate_values_(self) -> Tuple[str, cirq.PeriodicValue]:
-        return self.polarity, cirq.PeriodicValue(self.sandwich_rx_rads, 4 * np.pi)
+    def _value_equality_approximate_values_(self) -> Tuple[cirq.PeriodicValue, cirq.PeriodicValue]:
+        return (
+            cirq.PeriodicValue(self.theta, 2 * np.pi),
```
richrines1, line 328:
```suggestion
            cirq.PeriodicValue(self.theta, 4 * np.pi),
```

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -246,71 +254,111 @@ def _num_qubits_(self) -> int:
...
+        return approx_eq_mod(self.theta, other.theta, 2 * np.pi, atol=atol) and approx_eq_mod(
+            self.sandwich_rx_rads, other.sandwich_rx_rads, 2 * np.pi, atol=atol
+        )
 
     def __repr__(self) -> str:
-        if not self.sandwich_rx_rads:
-            return f"css.AceCR({self.polarity!r})"
-        return f"css.AceCR({self.polarity!r}, {self.sandwich_rx_rads!r})"
+        if not self.sandwich_rx_rads and self.theta == np.pi / 2:
+            return "css.AceCR"
```
richrines1, line 342: this should be
```suggestion
            return "css.AceCR()"
```
so that `eval(repr(gate)) == gate`

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -246,71 +254,111 @@ def _num_qubits_(self) -> int:
...
+            return f"AceCR({self.theta})"
+        elif self.theta == np.pi / 2:
+            return f"AceCR|{cirq.rx(self.sandwich_rx_rads)}|"
+        return f"AceCR({self.theta})|{cirq.rx(self.sandwich_rx_rads)}|"
 
     def _json_dict_(self) -> Dict[str, Any]:
-        return cirq.obj_to_dict_helper(self, ["polarity", "sandwich_rx_rads"])
+        if self.theta != np.pi / 2:
+            return cirq.obj_to_dict_helper(self, ["sandwich_rx_rads", "theta"])
+        return cirq.obj_to_dict_helper(self, ["sandwich_rx_rads"])
```
richrines1, line 361: no need for the special case here
```suggestion
        return cirq.obj_to_dict_helper(self, ["sandwich_rx_rads", "theta"])
```

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -711,23 +759,15 @@ def _from_json_dict_(cls, **kwargs: Any) -> "IXGate":
...
-        return cirq.ops.MSGate
-    if cirq_type == "RGate":
-        return RGate
-    if cirq_type == "IXGate":
-        return IXGate
-    if cirq_type == "ParallelRGate":
-        return ParallelRGate
-
-    return None
+    TYPE_TO_GATE_MAP: Dict[str, Type[cirq.Gate]] = {
```
richrines1, line 762: nice! though imo this shouldn't be capitalized if it's not global
```suggestion
    type_to_gate_map: Dict[str, Type[cirq.Gate]] = {
```

cirq_superstaq/ops/qubit_gates_test.py
```diff
@@ -229,57 +263,125 @@ def test_acecr_eq() -> None:
...
+    )
+    assert cirq.equal_up_to_global_phase(
+        css.AceCR(sandwich_rx_rads=np.pi), css.AceCR(sandwich_rx_rads=3 * np.pi)
+    )
 
     assert not cirq.equal_up_to_global_phase(css.AceCR("+-"), cirq.CX)
+    assert not cirq.equal_up_to_global_phase(css.AceCR(theta=np.pi), cirq.CX)
 
 
 def test_acecr_parameterized() -> None:
```
richrines1, line 329: a lot of this test could be outsourced `cirq.testing.assert_consistent_resolve_parameters`
```
for theta in (1.23, x):
    for sandwich_rx_rads in (0, 2.34, x, y):
        cirq.testing.assert_consistent_resolve_parameters(css.AceCR(theta, sandwich_rx_rads))

File: cirq_superstaq/ops/qubit_gates_test.py [hunk 7]
```diff
@@ -229,57 +263,125 @@ def test_acecr_eq() -> None:
...
+        repr(css.AceCR("+-", sandwich_rx_rads=np.pi))
+        == "css.AceCR(sandwich_rx_rads=3.141592653589793)"
+    )
+    assert repr(css.AceCR(theta=np.pi)) == "css.AceCR(theta=3.141592653589793)"
+    assert (
+        repr(css.AceCR(theta=np.pi / 3, sandwich_rx_rads=np.pi / 2))
+        == "css.AceCR(theta=1.0471975511965976, sandwich_rx_rads=1.5707963267948966)"
+    )
     cirq.testing.assert_equivalent_repr(
         css.AceCRMinusPlus, setup_code="import cirq_superstaq as css"
```
richrines1, line 375: imo there should be one of these for each special case in `__repr__` (in particular `assert_equicalent_repr(AceCRPlusMinus, ...)` would currently fail)

cirq_superstaq/ops/qubit_gates_test.py
```diff
@@ -229,57 +263,125 @@ def test_acecr_eq() -> None:
...
     )
-    assert str(css.AceCRMinusPlus) == "AceCR-+"
-    assert str(css.AceCR("+-", np.pi)) == "AceCR+-|Rx(π)|"
+    assert str(css.AceCR()) == "AceCR"
+    assert str(css.AceCR("+-", sandwich_rx_rads=np.pi)) == "AceCR|Rx(π)|"
+    assert str(css.AceCR(theta=np.pi)) == "AceCR(3.141592653589793)"
+    assert str(css.AceCR(theta=np.pi, sandwich_rx_rads=np.pi)) == "AceCR(3.141592653589793)|Rx(π)|"
 
 
 def test_acecr_decompose() -> None:
```
richrines1, line 387: this can also call `cirq.testing.assert_decompose_is_consistent_with_unitary()`

cirq_superstaq/ops/qubit_gates_test.py
```diff
@@ -191,21 +195,48 @@ def test_acecr_circuit_diagram_info() -> None:
...
 def test_acecr_qasm() -> None:
     qubits = cirq.LineQubit.range(2)
     circuit = cirq.Circuit(
         css.AceCR("+-").on(*qubits),
         css.AceCR("-+").on(*reversed(qubits)),
-        css.AceCR("-+", np.pi / 2).on(*qubits),
+        css.AceCR("+-", sandwich_rx_rads=np.pi / 2).on(*qubits),
+        css.AceCR("-+", sandwich_rx_rads=np.pi / 2).on(*qubits),
+        css.AceCR(theta=np.pi / 3).on(*qubits),
+        css.AceCR(theta=np.pi / 3, sandwich_rx_rads=np.pi / 2).on(*qubits),
```
richrines1, line 239: nit: can we use a different denominator here, e.g.
```suggestion
        css.AceCR(theta=np.pi / 5).on(*qubits),
        css.AceCR(theta=np.pi / 5, sandwich_rx_rads=np.pi / 2).on(*qubits),
```
so that the strings below don't depend on the default precision of the printer?

cirq_superstaq/ops/qubit_gates_test.py
```diff
@@ -218,9 +249,12 @@ def test_acecr_qasm() -> None:
...
 
-        acecr_pm q[0],q[1];
-        acecr_mp q[1],q[0];
-        acecr_mp_rx(pi*0.5) q[0],q[1];
+        acecr q[0],q[1];
+        acecr_rzx(pi*-0.5) q[1],q[0];
+        acecr_rx(pi*0.5) q[0],q[1];
+        acecr_rzx_rx(pi*-0.5,pi*0.5) q[0],q[1];
+        acecr_rzx(pi*0.3333333333) q[0],q[1];
+        acecr_rzx_rx(pi*0.3333333333,pi*0.5) q[0],q[1];
```
richrines1, line 257: (assuming prior suggestion)
```suggestion
        acecr_rzx(pi*0.2) q[0],q[1];
        acecr_rzx_rx(pi*0.2,pi*0.5) q[0],q[1];
```

richrines1 (approved):
looks good! bunch of small comments/suggestions

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -246,71 +254,111 @@ def _num_qubits_(self) -> int:
...
+        elif self.theta == -np.pi / 2:
+            theta_str = "-+"
+        else:
+            theta_str = f"({args.format_radians(self.theta)})"
+        top, bottom = (f"AceCR{theta_str}(Z side)", f"AceCR{theta_str}(X side)")
         if self.sandwich_rx_rads:
             bottom += f"|Rx({args.format_radians(self.sandwich_rx_rads)})|"
         return cirq.CircuitDiagramInfo(wire_symbols=(top, bottom))
 
     def _qasm_(self, args: cirq.QasmArgs, qubits: Tuple[cirq.Qid, cirq.Qid]) -> Optional[str]:
```
richrines1, line 276: wdyt about simplifying this so that the qasm is just `acecr(theta)` or `acecr_rx(theta, sandwich)`? it doesn't seem like we need unique qasm strings for every combination of default/non-default values

we could also use `ecr q[0],q[1]` for the default (`AceCRPlusMinus`) gate. this would be in keeping w/ qiskit's `ECRGate()`, however perhaps confusing in comparison to the other versions

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -246,71 +254,111 @@ def _num_qubits_(self) -> int:
...
+            theta=cirq.resolve_parameters(self.theta, resolver, recursive),
+            sandwich_rx_rads=cirq.resolve_parameters(self.sandwich_rx_rads, resolver, recursive),
         )
 
-    def _value_equality_values_(self) -> Tuple[str, cirq.TParamVal]:
+    def _value_equality_values_(self) -> Tuple[cirq.TParamVal, cirq.TParamVal]:
+        if cirq.is_parameterized(self.theta):
+            theta = self.theta
+        else:
+            theta = self.theta % (2 * np.pi)
```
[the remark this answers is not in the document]
dowusu-antwi: To clarify, if `2*pi` adds a global phase then shouldn't we want

```python
assert css.AceCR(rads=np.pi) == css.AceCR(rads=3 * np.pi)
```

to pass? Leaving the above as `theta = self.theta % (2 * np.pi)` allows it to pass (which I thought made sense because we're saying the `AceCR(theta)` gates should be equal up to a `2*pi` global phase on `theta`...)

It also allows

```python
assert css.AceCR(rads=np.pi) == css.AceCR(rads=-np.pi)
```

to pass, which I think should happen (polarity reversals are equal up to global phase).

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -230,14 +230,18 @@ class AceCR(cirq.Gate):
 
     Args:
         polarity: Should be either "+-" or "-+". Specifies if positive or negative half-CR is first
+        theta: Angle of rotation for CR gate (i.e., twice the angle for each echoed half-CR).
```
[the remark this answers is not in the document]
dowusu-antwi: Done.

cirq_superstaq/ops/qubit_gates_test.py
dowusu-antwi: Done.

cirq_superstaq/ops/qubit_gates.py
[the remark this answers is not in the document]
richrines1: > To clarify, if `2*pi` adds a global phase then shouldn't we want
> 
> ```python
> assert css.AceCR(rads=np.pi) == css.AceCR(rads=3 * np.pi)
> ```
> 
> to pass?

right --- `gate1 == gate2` (and `cirq.approx_eq(gate1, gate2)`) should only be true if gate1 and gate2 have the same global phase, while `cirq.equal_up_to_global_phase(gate1, gate2)` should be true either way. This is how all cirq builtin gates work, e.g.:
```python
assert cirq.rx(np.pi) == cirq.rx(-np.pi)  # fails
assert cirq.rx(np.pi) == cirq.rx(3 * np.pi)  # fails
assert cirq.rx(np.pi) == cirq.rx(5 * np.pi)  # succeeds

assert cirq.approx_eq(cirq.rx(np.pi), cirq.rx(-np.pi))  # fails
assert cirq.approx_eq(cirq.rx(np.pi), cirq.rx(3 * np.pi))  # fails
assert cirq.approx_eq(cirq.rx(np.pi), cirq.rx(5 * np.pi))  # succeeds

assert cirq.equal_up_to_global_phase(cirq.rx(np.pi), cirq.rx(-np.pi))  # succeeds
assert cirq.equal_up_to_global_phase(cirq.rx(np.pi), cirq.rx(3 * np.pi))  # succeeds
assert cirq.equal_up_to_global_phase(cirq.rx(np.pi), cirq.rx(5 * np.pi))  # succeeds
```

cirq_superstaq/ops/qubit_gates.py
dowusu-antwi: Simplified to `acecr(rads)` / `acecr_rx(rads, sandwich)`.

`AceCRPlusMinus` (i.e., `AceCR(np.pi / 2)`) becomes `acecr(pi*0.5)`. Shouldn't qasm distinguish between qiskit's `ECRGate` and our `AceCR(np.pi / 2)`?

cirq_superstaq/ops/qubit_gates.py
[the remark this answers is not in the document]
dowusu-antwi: Done.

cirq_superstaq/ops/qubit_gates.py
richrines1: > `AceCRPlusMinus` (i.e., `AceCR(np.pi / 2)`) becomes `acecr(pi*0.5)`. Shouldn't qasm distinguish between qiskit's `ECRGate` and our `AceCR(np.pi / 2)`?

ok i agree, i think it's definitely better to keep it the way you have it (with `acecr(pi*0.5)`). but just out of curiosity - is `AceCR(pi/2)` different from `ECRGate()`?

dowusu-antwi: Oh good point. I'm thinking of the single-qubit gate optimizations we do around the `AceCR` in the `CNOT` gate -- `AceCR(pi/2)` and `ECRGate` are identical.

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -246,71 +254,104 @@ def _num_qubits_(self) -> int:
...
+            return "AceCR"
+        elif not self.sandwich_rx_rads:
+            return f"AceCR({self.rads})"
+        elif self.rads == np.pi / 2:
+            return f"AceCR|{cirq.rx(self.sandwich_rx_rads)}|"
+        return f"AceCR({self.rads})|{cirq.rx(self.sandwich_rx_rads)}|"
 
     def _json_dict_(self) -> Dict[str, Any]:
-        return cirq.obj_to_dict_helper(self, ["polarity", "sandwich_rx_rads"])
+        return cirq.obj_to_dict_helper(self, ["sandwich_rx_rads", "rads"])
```
richrines1, line 354: nit
```suggestion
        return cirq.obj_to_dict_helper(self, ["rads", "sandwich_rx_rads"])
```

richrines1 (commented):
a few more nits then lgtm!

richrines1 (approved):
lgtm!

dowusu-antwi:
> lgtm!

Thanks for the review!

PR #427, opened 2023-02-01
Account for delays on drive channel

dowusu-antwi:
Control channel `U0` is logical -- i.e., it's just a label, and physically coincides with drive channel `D0`. To avoid overlapping pulses, IBM includes a drive channel delay per control channel pulse in echoed `RZX`.

Why `5 -> 7` in file changed? We didn't have these 2 additional delay instructions before because we extracted CR halves from CNOT:

download

(This fixes recent `deploy_integration_test` server-side failure.)
image


richrines1 (approved):
lgtm! (just tested locally)

PR #430, opened 2023-02-01
add qubit_subspace_op

richrines1:
(just a helper for constructing ops with `QubitSubspaceGate`)

also makes BSwapPowGate an `InterchangeableQubitsGate` (should have been done in #420)


paaige (approved):
lgtm

PR #431, opened 2023-02-03
fix diagram of CZ3_INV

richrines1:
(before the exponent was being left out, so e.g. CZ3_INV and CZ3 would have the same diagram)


singular-value (approved):
lgtm

PR #434, opened 2023-02-07
Workflows for automated releases

stephanielee9:
Updated / added three files:

1. `main.yaml` - added a job to release after a successful deploy on main
2. `ci.yml` - added workflow_call (so it can be used by other workflows) and ignored updates to `_version.py` (so the action isn't run twice - it'll already be called by the new workflow below)
3. `automated_publish_release.yml` - once `_version.py` is pushed to main, will run checks, if they pass will then run the `main.yaml` script. If the checks or deploy fails, someone can manually run `main.yaml` to try to deploy and publish release again.


.github/workflows/ci.yml
```diff
@@ -6,8 +6,11 @@ name: cirq-superstaq
 on:
   push:
     branches: [main]
+    paths-ignore:
```
dowusu-antwi, line 9: This is just to avoid running the checks twice right?

.github/workflows/ci.yml
```diff
@@ -6,8 +6,11 @@ name: cirq-superstaq
...
+      - 'cirq_superstaq/_version.py'
   pull_request:
     branches: [main]
+  workflow_call: null
```
dowusu-antwi, line 13: What is this line for? (Same question for line 5 in `main.yaml`)

stephanielee9: `workflow_call` allows the workflow to be called by other workflows. In this case, the automated release workflow (`automated_publish_release.yml`) is just a workflow triggered when updates to `_version.py` are made to run the checks workflow and then the deploy and release workflow, if the checks pass.

dowusu-antwi: Makes sense, thanks!

.github/workflows/automated_publish_release.yml
```diff
@@ -0,0 +1,22 @@
...
+      - 'cirq_superstaq/_version.py'
+
+jobs:
+  run_checks:
+    uses: ./.github/workflows/ci.yml
+    secrets: inherit
+  publish_and_release:
+    needs: run_checks
+    uses: ./.github/workflows/main.yaml
+    secrets: inherit
```
dowusu-antwi, line 22: @richrines1 @singular-value We've seen this before but I forgot how it was resolved -- adding `\n`? (The other `.yaml`s don't have it though...)

dowusu-antwi (dismissed):
lgtm, pending nit

.github/workflows/automated_publish_release.yml
stephanielee9: There is a more sophisticated way to do this that I think @vtomole knows about, but I usually just go in GitHub and manually enter a new line (hit Enter key, delete any auto-indents GitHub adds).

dowusu-antwi: Fixed now -- thanks @stephanielee9!

dowusu-antwi (approved):

PR #436, opened 2023-02-09
`css.Barrier(n) != cirq.IdentityGate(n)`

richrines1:
currently `css.Barrier(n) == cirq.IdentityGate(n)` is True, but it shouldn't be (they have the same unitary but definitely aren't interchangeable)

`isinstance(css.Barrier(n), cirq.IdentityGate)` will remain True


perlinm (approved):
nice and simple.  lgtm.

PR #438, opened 2023-02-13
add gate_defs argument to aqt_compile

richrines1:


cirq_superstaq/service.py
```diff
@@ -25,6 +25,25 @@
 import cirq_superstaq as css
 
 
+def _to_matrix_gate(matrix: npt.ArrayLike) -> cirq.MatrixGate:
+    """Convert a unitary matrix into a cirq.MatrixGate. If the matrix's dimension is a power of 3,
+    assume to be a qutrit gate; otherwise, assume qubits.
+
+    Args:
+        matrix: The (unitary) matrix to be converted.
+
```
paaige, line 34: would it be overkill to replace 3 with "qutrit_base" or something here? also replacing 9. I feel like we don't normally have hardcoded numbers in non test code and it's sometimes easier to read, even if it seems very obvious what the numbers represent

cirq_superstaq/compiler_output.py
```diff
@@ -132,7 +132,7 @@ def read_json_ibmq(json_dict: Dict[str, Any], circuits_is_list: bool) -> Compile
 
 
 def read_json_aqt(  # pylint: disable=missing-param-doc
-    json_dict: Dict[str, Any], circuits_is_list: bool, num_eca_circuits: int = 0
+    json_dict: Dict[str, Any], circuits_is_list: bool, num_eca_circuits: Optional[int] = None
```
paaige, line 135: Why this change?

richrines1: `_aqt_compile()` always passes a `num_eca_circuits` argument to `read_json_aqt()` regardless of whether it's called from `aqt_compile()` and `aqt_compile_eca()`, so it seemed a lot more clear to have it pass the default `None` for non-eca compilation instead of having it pass 0 and treating that as a kind of magic number

(your comment also made me realize i forgot to propagate this change through the rest of the function; fixed now)

cirq_superstaq/service.py
richrines1: just refactored a bit to replace the hardcoded values with a more generic `dimension` variable

cirq_superstaq/compiler_output.py
richrines1: though now i'm thinking we should also raise an error in `aqt_compile_eca` if you pass `num_equivalent_circuits=0`

richrines1: > though now i'm thinking we should also raise an error in `aqt_compile_eca` if you pass `num_equivalent_circuits=0`

will do this in a follow-up pr

paaige (approved):
lgtm

cirq_superstaq/compiler_output.py
richrines1: just filed https://github.com/SupertechLabs/general-superstaq/issues/227 (with universal tag)

PR #440, opened 2023-02-14
Add qscout_tutorial.ipynb to cirq-superstaq/examples 

hallben4:


richrines1:
notebooks in this repo should only include cirq examples - the qiskit versions should be in qss/examples (c.f. examples/aqt.ipynb, which is in both repos with ~the same set of examples, but implemented with cirq in css and qiskit in qss)

pnoell:
Remove import qss from notebook

PR #441, opened 2023-02-15
New Stripped CZ Gate

SalahedeenIssa:
Fixes https://github.com/SupertechLabs/cirq-superstaq/issues/412


cirq_superstaq/ops/qubit_gates_test.py
```diff
@@ -996,7 +1059,7 @@ def test_custom_resolver() -> None:
     circuit += css.AQTITOFFOLI(qubits[0], qubits[1], qubits[2])
     circuit += css.ops.qubit_gates.ICCX(qubits[0], qubits[1], qubits[2])
     circuit += css.ops.qubit_gates.IX(qubits[0])
-    circuit += cirq.CX(qubits[0], qubits[1])
```
paaige, line 999: just a question (for understanding), why did you remove this line?

SalahedeenIssa: Honestly @perlinm and I just thought it was redundant, so we removed it and checked to see if it made any difference in the checks and it didn't. So no real reason besides that.

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -753,6 +753,75 @@ def _from_json_dict_(cls, **kwargs: Any) -> "IXGate":
...
+
+    def _parameter_names_(self) -> AbstractSet[str]:
+        return cirq.parameter_names(self.phi)
+
+    def _has_unitary_(self) -> bool:
+        return not self._is_parameterized_()
+
+    def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.CircuitDiagramInfo:
+        p = args.format_radians(self.phi)
+        return cirq.CircuitDiagramInfo(wire_symbols=("@", f"@({p})"))
```
perlinm, line 819: nit (previously discussed offline, but adding here so we don't forget):
```suggestion
        return cirq.CircuitDiagramInfo(wire_symbols=("@({p})", f"@({p})"))
```
It's technically duplicate info, but it signals that both qubits are modified by an Rz gate.

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -753,6 +753,75 @@ def _from_json_dict_(cls, **kwargs: Any) -> "IXGate":
...
+        if cirq.is_parameterized(self.phi):
+            return self.phi
+        return self.phi % (2 * np.pi)
+
+    def _value_equality_approximate_values_(self) -> cirq.PeriodicValue:
+        return cirq.PeriodicValue(self.phi, 2 * np.pi)
+
+    def __pow__(self, exponent: cirq.TParamVal) -> "StrippedCZGate":
+        if exponent == 1:
+            return StrippedCZGate(exponent * self.phi)
```
richrines1, line 787: is it worth adding the inverse here? we can use:
```suggestion
        if exponent == 1:
            return StrippedCZGate(self.phi)
        elif exponent == -1:
            return StrippedCZGate(-self.phi)
```
(also removed the `exponent *` because the exponent is 1)

cirq_superstaq/ops/qubit_gates_test.py
```diff
@@ -108,6 +108,69 @@ def test_zz_swap_parameterized() -> None:
...
+        cirq.CZ(qubits[0], qubits[2]),
+    ]
+    cirq.testing.assert_has_consistent_apply_unitary(gate)
+    cirq.testing.assert_decompose_is_consistent_with_unitary(gate, ignoring_global_phase=True)
+    cirq.testing.assert_consistent_resolve_parameters(gate)
+
+    assert cirq.approx_eq(css.StrippedCZGate(1.23), css.StrippedCZGate(1.23 + 2 * np.pi))
+    assert cirq.approx_eq(css.StrippedCZGate(np.pi + 1e-10), css.StrippedCZGate(np.pi - 1e-10))
+
+    assert gate**1 == gate
```
richrines1, line 136:
```suggestion
    assert gate**1 == gate
    assert gate**-1 == css.StrippedCZGate(-gate.phi)
```
(assuming prior suggestion)

cirq_superstaq/ops/qubit_gates_test.py
```diff
@@ -108,6 +108,69 @@ def test_zz_swap_parameterized() -> None:
...
+    qubits = cirq.LineQubit.range(3)
+    operation = gate(qubits[0], qubits[2])
+    assert cirq.decompose_once(operation) == [
+        cirq.rz(phi).on(qubits[0]),
+        cirq.rz(phi).on(qubits[2]),
+        cirq.CZ(qubits[0], qubits[2]),
+    ]
+    cirq.testing.assert_has_consistent_apply_unitary(gate)
+    cirq.testing.assert_decompose_is_consistent_with_unitary(gate, ignoring_global_phase=True)
+    cirq.testing.assert_consistent_resolve_parameters(gate)
```
richrines1, line 131: we can also add:
```suggestion
    cirq.testing.assert_consistent_resolve_parameters(gate)
    cirq.testing.assert_json_roundtrip_works(
        gate, resolvers=[*css.SUPERSTAQ_RESOLVERS, *cirq.DEFAULT_RESOLVERS]
    )
```

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -753,6 +753,75 @@ def _from_json_dict_(cls, **kwargs: Any) -> "IXGate":
...
 
 
+@cirq.value_equality(approximate=True)
+class StrippedCZGate(cirq.Gate):
+    """The Stripped CZ Gate is the gate that is actually being performed by Hilbert,
+    which is just a regular CZ gate when phi = 0, and is corrected into a CZ
+    gate by RZ gates afterwards if phi is nonzero.
+    """
+
+    def __init__(self, phi: cirq.TParamVal = 0) -> None:
```
richrines1, line 763: can we use a more descriptive variable name? e.g. something like
```suggestion
    def __init__(self, rz_rads: cirq.TParamVal = 0) -> None:
```

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -753,6 +753,75 @@ def _from_json_dict_(cls, **kwargs: Any) -> "IXGate":
...
+    def __pow__(self, exponent: cirq.TParamVal) -> "StrippedCZGate":
+        if exponent == 1:
+            return StrippedCZGate(exponent * self.phi)
+        return NotImplemented
+
+    def __str__(self) -> str:
+        return f"StrippedCZGate({self.phi})"
+
+    def __repr__(self) -> str:
+        return f"css.StrippedCZGate({self.phi})"
```
richrines1, line 794:
```suggestion
        return f"css.StrippedCZGate({self.phi!r})"
```

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -753,6 +753,75 @@ def _from_json_dict_(cls, **kwargs: Any) -> "IXGate":
...
+@cirq.value_equality(approximate=True)
+class StrippedCZGate(cirq.Gate):
+    """The Stripped CZ Gate is the gate that is actually being performed by Hilbert,
+    which is just a regular CZ gate when phi = 0, and is corrected into a CZ
+    gate by RZ gates afterwards if phi is nonzero.
+    """
+
+    def __init__(self, phi: cirq.TParamVal = 0) -> None:
+        """Args: phi: CZ-rotation angle in radians"""
+        self.phi = phi
```
richrines1, line 765: nit
```suggestion
        self._phi = phi

    @property
    def phi(self) -> cirq.TParamVal:
        return self._phi
```
(or something else in place of `phi`, as per prior comment)

cirq_superstaq/ops/qubit_gates.py
```diff
@@ -753,6 +753,75 @@ def _from_json_dict_(cls, **kwargs: Any) -> "IXGate":
...
 
+@cirq.value_equality(approximate=True)
+class StrippedCZGate(cirq.Gate):
+    """The Stripped CZ Gate is the gate that is actually being performed by Hilbert,
+    which is just a regular CZ gate when phi = 0, and is corrected into a CZ
+    gate by RZ gates afterwards if phi is nonzero.
+    """
+
+    def __init__(self, phi: cirq.TParamVal = 0) -> None:
+        """Args: phi: CZ-rotation angle in radians"""
```
richrines1, line 764: this docstring should be updated - phi is changing the angle of the single-qubit rzs, not the CZ rotation angle

cirq_superstaq/ops/qubit_gates.py
perlinm: We may as well do arbitrary integer exponents at this point, with something like
```python
if exponent == 0:
    return cirq.IdentityGate(2)
elif exponent == int(exponent):
    new_phi = exponent * phi
    if exponent % 2 == 1:
        return cirq.StrippedCZGate(new_phi)
    return cirq.ParallelGates(cirq.Rz(new_phi), cirq.Rz(new_phi))
return NotImplemented
```

perlinm: Also, is `Gate.__pow__` allowed to return a sequence of gates?  If so, we could also do non-integer powers.

perlinm: Actually: is `Gate.__pow__` allowed to return a sequence of gates?  If so, maybe we should return a `CZPowGate` + `ParallelGates` for non-integer exponents.

richrines1: nice. to preserve global phase in the fractional/DiagonalGate case i think it has to be (definitely double-check):
```python
phases = [np.exp(-1j * new_phi), 1, 1, np.exp(1j * (new_phi - exponent * np.pi))]
```
though i might also be tempted to skip this case and only support integer powers

(also i think multiple-of-two case should be `cirq.ParallelGate(cirq.rz(new_phi), 2)`)

richrines1: > Actually: is `Gate.__pow__` allowed to return a sequence of gates?

i don't think so

perlinm: For global phase, the phases I had are correct: the |00> state is completely unaffected by the gate.  So actually the `Rz`s I had should be `ZPowGate`s.

richrines1: out of curiosity what do you mean by "correct"? physically aren't the two gates the same? (obviously nothing wrong with with the ZPowGate version, just interested)

richrines1: oh wait does it have to do with the rydberg level?

perlinm: I meant "correct" in the sense of getting the global phase correct as well.  The |00> state is completely unaffected by the pulses that implement the CZ gate (which couples only the |1> state to the rydberg level), so the <00|gate|00> should be 1.

cirq_superstaq/ops/qubit_gates.py
perlinm: I would support making that both here and everywhere else in the class.

Follow-up proposal: maybe we should make this parameter something like `z_exponent` and make it a multiple of `pi`?

SalahedeenIssa: I think I'll do the rz_rads for now, and then if/when we change ParallelRGates to the multiples of pi format I'll change it here too.

cirq_superstaq/ops/qubit_gates.py
perlinm: @SalahedeenIssa please comment something like "done" when you fix these, so we know that you've seen + addressed it

perlinm (dismissed):
LGTM!

Next up is doing the same in `qiskit-superstaq`

perlinm (approved):

PR #444, opened 2023-02-16
bump version to 0.3.18

richrines1:


dowusu-antwi (approved):
LGTM.

dowusu-antwi:
Is there sometimes a slight lag between publishing a PyPi package and being able to  `pip install` it? I think that's why tests are failing here.

richrines1:
> Is there sometimes a slight lag between publishing a PyPi package and being able to `pip install` it?

yeah i think i was too impatient, it seems to take a few minutes usually

PR #446, opened 2023-02-28
add final-logical-to-physical to qscout_compile

richrines1:
(most important for qscout because of swap mirroring, but will eventually be added to all compile endpoints)


hallben4 (approved):
Looks great!

PR #451, opened 2023-03-07
prevent clobbering of job_id when job is refreshed

richrines1:
the id of a given job should be fixed, so it shouldn't be part of the `job._job` dictionary (which gets updated whenever the job is refreshed)

(no need to fix in `qss.superstaq_job`, which already prevents updates from modifying `job_id`s)

Fixes https://github.com/SupertechLabs/cirq-superstaq/issues/452


vtomole (approved):
LGTM

vtomole:
Thoughts on adding

```python
qubit = cirq.LineQubit(0)
circuit = cirq.Circuit(
    cirq.X(qubit),
    cirq.measure(qubit),
    )

job = service.create_job(circuit=circuit, repetitions=1, target="ibmq_qasm_simulator")
assert job.counts() == {'1': 1}
```

As a daily integration test in this PR?

richrines1:
good call, done. (though had to structure it a little differently to make sure it's tested before and after being refreshed)

also probably good just to have an integration test which actually creates a job (which somehow we didn't before?)

vtomole (approved):
> which somehow we didn't before?

We didn't do this because we didn't want to flood the cloud with bot jobs. But this is daily so it's not too bad and now also makes sense given our upcoming DB garbage collector

CC @dowusu-antwi @PalashGoiporia and @bharat-thotakura for awareness about this.

PR #456, opened 2023-03-13
Allow reading SuperstaQ API key from an authentication file

perlinm:
If a SuperstaQ API key is not passed to `css.Service` and not set as an environment variable, try to read it from a file.


cirq_superstaq/service.py
```diff
@@ -559,3 +568,15 @@ def supercheq(
...
         fidelities = gss.serialization.deserialize(json_dict["fidelities"])
         return circuits, fidelities
+
+
+def _find_api_key() -> Optional[str]:
+    """Try to load a SuperstaQ API key from a config file."""
+
+    for directory in ["~/.config", "~/.local/share", "~/.super.tech", "~/.coldquanta"]:
+        path = os.path.join(*directory.split("/"), "superstaq_api_key")
+        if os.path.isfile(path):
```
richrines1, line 578: a few note here:
1. `~/.config` and `~/.local/share` are the default/most commonly used xdg spec directories, but technically they're supposed to be configurable via the `$XDG_CONFIG_HOME` and `$XDG_DATA_HOME` variables
2. we can probably skip the former - usually credentials are supposed to be stored in the xdg_data_home i think
3. in the latter, the key file should still be in either a `super.tech` or `coldquanta` subdirectory (but with no leading `.`)
4. in all cases, we should be using `expanduser()` to properly expand the `~`
5. these are common posix paths - are there similar standards in windows?
```suggestion
    data_dir = os.getenv("XDG_DATA_HOME", "~/.local/share")
    for directory in [
        os.path.join(data_dir, "super.tech"),
        os.path.join(data_dir, "coldquanta"),
        "~/.super.tech",
        "~/.coldquanta",
    ]:
        path = os.path.expanduser(os.path.join(*directory.split("/"), "superstaq_api_key"))
        if os.path.isfile():
```
or alternatively using `pathlib` to handle all the expansion/joining/etc:
```suggestion
    data_dir = pathlib.Path(os.getenv("XDG_DATA_HOME", "~/.local/share")).expanduser()
    home_dir = pathlib.Path.home()

    for path in [
        data_dir.joinpath("super.tech", "superstaq_api_key"),
        data_dir.joinpath("coldquanta", "superstaq_api_key"),
        home_dir.joinpath(".super.tech", "superstaq_api_key"),
        home_dir.joinpath(".coldquanta", "superstaq_api_key"),
    ]:
        if path.is_file():
```

richrines1: also wdyt about just using "api_key" for the filename when it's in one of the super.tech directories? the "superstaq_" part is maybe redundant but can't decide if this would be nice or just overcomplicated

cirq_superstaq/service.py
```diff
@@ -559,3 +568,15 @@ def supercheq(
         circuits = css.serialization.deserialize_circuits(json_dict["cirq_circuits"])
         fidelities = gss.serialization.deserialize(json_dict["fidelities"])
         return circuits, fidelities
+
+
+def _find_api_key() -> Optional[str]:
```
richrines1, line 573: this should probably live in `gss`

cirq_superstaq/service.py
perlinm: I like the `pathlib` suggestion, thanks!

The only issue with naming the file `api_key` is that there is a separate CQCloud API key in `~/.coldquanta/cq_auth.json`, so that can be a source of confusion.

cirq_superstaq/service.py
perlinm: Ah good point.

PR #459, opened 2023-03-24
link to new home in readme

richrines1:
part of https://github.com/SupertechLabs/client-superstaq/issues/397

also disables the automated daily integration check (which is now run from the new repo)


richrines1:
@vtomole the checks won't pass on this until we update the gss requirement - do you think it's better to override the check or to update? (it feels wrong to check requirements.txt if we want to keep this repo as a snapshot of css's state when we migrated)

vtomole:
Let's override the check.

vtomole:
Disabled. Merging.

